PR #2, opened 2017-10-12
Changes to work with roachperf.

mberhault:
* store username/ip address in hosts file
* GC hosts files
* rework internal datastructures a bit
* more details in `roachprod list`
* reduce default lifetime to 12h
* indexing at 1 to match crl-prod/roachperf
* readme changes

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -14,6 +14,7 @@ type Cloud struct {
 	// Individual "bad" instances by category.
 	InvalidName  []jsonVM
 	NoExpiration []jsonVM
+	BadNetwork   []jsonVM
 }
 
 func newCloud() *Cloud {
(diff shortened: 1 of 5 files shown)


mberhault:
More changes added. Mainly so that it can be used by `roachperf`. see https://github.com/cockroachdb/roachperf/pull/5

cluster.go
```diff
@@ -115,7 +150,8 @@ func listCloud() (*Cloud, error) {
 func createCluster(name string, nodes int, lifetime time.Duration) error {
 	vmNames := make([]string, nodes, nodes)
 	for i := 0; i < nodes; i++ {
-		vmNames[i] = fmt.Sprintf("%s-%0.4d", name, i)
+		// Start instance indexing at 1.
```
petermattis: Alternately, we could change `roachperf` and other tools as well.

mberhault: The original reason to move from 0-indexing to 1-indexing was to have hostnames match node IDs (when nodes are brought up in sequential order).
I think this is still desirable. If the nodes are brought up in machine-name order (difficult with concurrency, but we could tweak things to make it happen again), it's still less misleading to have it match.

PR #4, opened 2017-10-16
Use local SSDs and format using startup script.

mberhault:
Other tweaks:
* reduce boot disk size (50G -> 10G)
* fix name parsing

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -70,10 +70,10 @@ func (vl VMList) Less(i, j int) bool { return vl[i].Name < vl[j].Name }
 
 func namesFromVMName(name string) (string, string, error) {
 	parts := strings.Split(name, "-")
-	if len(parts) != 3 {
+	if len(parts) < 3 {
 		return "", "", fmt.Errorf("expected VM name in the form %s, got %s", vmNameFormat, name)
 	}
-	return parts[0], parts[0] + "-" + parts[1], nil
+	return parts[0], strings.Join(parts[:len(parts)-1], "-"), nil
 }
 
 func listCloud() (*Cloud, error) {
@@ -147,14 +147,14 @@ func listCloud() (*Cloud, error) {
 	return cloud, nil
 }
 
-func createCluster(name string, nodes int, lifetime time.Duration) error {
+func createCluster(name string, nodes int, opts VMOpts) error {
 	vmNames := make([]string, nodes, nodes)
 	for i := 0; i < nodes; i++ {
 		// Start instance indexing at 1.
 		vmNames[i] = fmt.Sprintf("%s-%0.4d", name, i+1)
 	}
 
-	return createVMs(vmNames, lifetime)
+	return createVMs(vmNames, opts)
 }
 
 func destroyCluster(c *Cluster) error {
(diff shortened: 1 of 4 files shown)


cluster.go
```diff
@@ -70,10 +70,10 @@ func (vl VMList) Less(i, j int) bool { return vl[i].Name < vl[j].Name }
 
 func namesFromVMName(name string) (string, string, error) {
 	parts := strings.Split(name, "-")
-	if len(parts) != 3 {
+	if len(parts) < 3 {
 		return "", "", fmt.Errorf("expected VM name in the form %s, got %s", vmNameFormat, name)
 	}
-	return parts[0], parts[0] + "-" + parts[1], nil
+	return parts[0], strings.Join(parts[0:len(parts)-1], "-"), nil
```
petermattis: FYI, the `0:` isn't necessary. This could be `strings.Join(parts[:len(parts)-1], "-")`.

utils.go
```diff
@@ -0,0 +1,51 @@
...
+func writeStartupScript() (string, error) {
+	tmpfile, err := ioutil.TempFile("", "gce-startup-script")
+	if err != nil {
+		return "", err
+	}
+
+	if _, err := tmpfile.WriteString(gceLocalSSDStartupScript); err != nil {
+		return "", err
+	}
+	if err := tmpfile.Close(); err != nil {
```
petermattis: The common pattern is to add a `defer tmpfile.Close()` immediately after opening the file.

utils.go
```diff
@@ -0,0 +1,51 @@
+package main
+
+import (
+	"io/ioutil"
+)
+
+// Startup script used to find/format/mount all local SSDs in GCE.
+// Each disk is mounted to /mnt/data<disknum> and chmoded to all users.
+const gceLocalSSDStartupScript = `
+#!/usr/bin/env bash
```
petermattis: I thought this had to be the first line of the file. This is making it the second.

gcloud.go
```diff
@@ -91,10 +99,13 @@ func createVMs(names []string, lifetime time.Duration) error {
 		"--scopes", "https://www.googleapis.com/auth/devstorage.read_only,https://www.googleapis.com/auth/logging.write,https://www.googleapis.com/auth/monitoring.write,https://www.googleapis.com/auth/servicecontrol,https://www.googleapis.com/auth/service.management.readonly,https://www.googleapis.com/auth/trace.append",
 		"--image", "ubuntu-1604-xenial-v20171002",
 		"--image-project", "ubuntu-os-cloud",
-		"--boot-disk-size", "50",
+		"--boot-disk-size", "10",
 		"--boot-disk-type", "pd-ssd",
+		"--local-ssd", "interface=SCSI",
```
petermattis: Did you want to make this optional?

cluster.go
mberhault: Done

gcloud.go
mberhault: Yup. Done, as well as `--machine-type`. Added a `VMOpts` to pass everything around.

utils.go
mberhault: Still works, but yes, you're right. Done.

utils.go
mberhault: Done.

PR #5, opened 2017-10-17
Monitor clusters and email.

mberhault:
* checks all clusters when run
* delete clusters more than 12h past the expiration date
* send email to users with deleted clusters and expired clusters
* don't keep emailing for expired clusters, just once. When emailing,
always email everything.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -12,17 +12,17 @@ const vmNameFormat = "user-<clusterid>-<nodeid>"
 type Cloud struct {
 	Clusters map[string]*Cluster
 	// Individual "bad" instances by category.
-	InvalidName  []jsonVM
-	NoExpiration []jsonVM
-	BadNetwork   []jsonVM
+	InvalidName  JsonVMList
+	NoExpiration JsonVMList
+	BadNetwork   JsonVMList
 }
 
 func newCloud() *Cloud {
 	return &Cloud{
 		Clusters:     make(map[string]*Cluster),
-		InvalidName:  make([]jsonVM, 0),
-		NoExpiration: make([]jsonVM, 0),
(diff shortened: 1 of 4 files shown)


mberhault:
odd. Github is not picking up the changes to the branch: https://github.com/cockroachlabs/roachprod/commit/d501ff96c63006e120cb05e614daddf424c8de98#diff-04c6e90faac2675aa89e2176d2eec7d8

README.md
```diff
@@ -83,7 +83,41 @@ OK
...
+  marc-test-0002	marc-test-0002.us-east1-b.cockroach-ephemeral	10.142.0.5	35.196.95.207
+  marc-test-0003	marc-test-0003.us-east1-b.cockroach-ephemeral	10.142.0.7	35.196.151.250
+  marc-test-0004	marc-test-0004.us-east1-b.cockroach-ephemeral	10.142.0.6	35.196.194.230
+Syncing...
+
+# Add gcloud SSH key
+$ ssh-add ~/.ssh/google_compute_engine
+
+# Stage scripts and binaries
+$ crl-stage-binaries marc-test all scripts
```
jordanlewis: It would be cool if we could roll these steps into `roachprod`. What's the point of a `roachprod` cluster that doesn't have a workable cockroach/loadgen setup on it?

Also, it would be useful to include the instructions to stage a local cockroach binary (`roachperf marc-test put cockroach`)

mberhault: moved to #6 and added the `put cockroach` example.
I agree that three tools (roachprod/crl-prod/roachperf) is getting unwieldy. My plan is to collapse it into at most 2, but that's for alter.

PR #13, opened 2017-10-28
Add extend command.

### LINKED ISSUE #12: Add ability to extend lifetime of cluster
For example, here is how to do it using `gcloud`. Having this integrated into `roachprod` would be a nice convenience. Apparently `add-labels` adds or updates labels. No need for a `remove-labels` first.

```
for i in $(seq 1 7); do 
  gcloud compute instances add-labels --project cockroach-ephemeral peter-perf-000$i \
    --labels lifetime=36h0m0s
done
```

mberhault:
Fixes #12.

Extends lifetime of all VMs in the cluster by a default of 12h
(`--lifetime`).

Unfortunately, gcloud has to be called on every instance, so it's a bit
slow.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -4,7 +4,10 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 	"time"
+
+	"github.com/pkg/errors"
 )
 
 const vmNameFormat = "user-<clusterid>-<nodeid>"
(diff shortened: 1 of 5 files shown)


cluster.go
```diff
@@ -166,3 +177,13 @@ func destroyCluster(c *Cluster) error {
 
 	return deleteVMs(vmNames)
 }
+
+func extendCluster(c *Cluster, extension time.Duration) error {
+	newLifetime := c.Lifetime + extension
+	for _, vm := range c.VMs {
+		if err := extendVM(vm.Name, newLifetime); err != nil {
```
petermattis: Does it help to do this in parallel? For example:

```
  var wg sync.WaitGroup
  wg.Add(len(c.VMs))
  for i := range c.VMs {
    go func(i int) {
      defer wg.Done()
      vm := c.VMs[i]
      ...
    }(i)
  }
  wg.Wait()
```

mberhault: Sude, done.

PR #20, opened 2017-12-15
Make tutorial commands more easily copy-pastable

a-robinson:
Having to edit the name in a copied command before running it bugs me.
Including the output for some of the commands but not all of them
doesn't really aid understanding when all of the commands are already
commented.
The shell prompts don't help clarity with the output removed, and they
make copying a little harder.

Just a minor pet peeve of mine.


mberhault:
LGTM. I should really drop the whole forced-user thing and let you specify it and just complain if you don't put your name. That would integrate with roachperf/crl-ssh better.

PR #21, opened 2018-01-24
Require full cluster names on the command line

### LINKED ISSUE #14: Use full cluster names
The roachprod command line uses truncated table names: the full cluster name is whatever is specified on the command line prefixed with `<username>-`. This make it awkward to switch between use `crl-*` and `roachperf` (which use full cluster names) and `roachprod` (which uses the truncated names). 

I propose that `roachprod create` takes a full cluster name and we require it is prefixed with `<username>-`. `roachprod {destroy,extend}` should take a full cluster name.

petermattis:
Verify cluster names are prefixed with the active account.

Fixes #14

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachlabs/roachprod/21)

### DIFF
--- a/main.go
+++ b/main.go
@@ -25,9 +25,8 @@ var (
 	destroyAfter   time.Duration
 	trackingFile   string
 	extendLifetime time.Duration
-    zones = ZoneList{"us-east1-b", "us-west1-b", "europe-west2-b"}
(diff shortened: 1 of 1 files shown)


petermattis:
There is definitely a nightly script in `roachperf` that needs fixing. I'll also send a PSA to #engineering.

PR #24, opened 2018-01-25
Store locality information in hosts files

benesch:
Store the locality of each node in the cluster's hosts file so that
roachperf can set the --locality flag appropriately when booting the
cluster.

See also: cockroachdb/roachperf#22, cockroachlabs/production#537

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"fmt"
+	"log"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


cluster.go
```diff
@@ -69,7 +70,17 @@ type VM struct {
...
 	PrivateIP string
 	PublicIP  string
-	Zone       string
+	Zone      string
+}
+
+var regionRE = regexp.MustCompile(`(.*[^-])-?[a-z]$`)
+
+func (vm *VM) locality() string {
+	region := "?"
```
petermattis: Perhaps region should be the empty string if there is no region? Not as obvious to the reader, but we almost never view this file.

benesch: The following isn't a valid locality:

```
--locality=region=,zone=some-zone
```

So: we could either omit it entirely (`--locality=zone=some-zone`), or return an error to the user. Do you have a preference? I have a slight preference for the latter. The more we can assert about our cluster's environment (e.g., "you will _always_ have a region and a zone"), the easier it is to write nightly tests.

petermattis: Ok, makes sense. Carry on.

benesch: Ok, roachprod now `log.Fatal`s if it can't parse a VM's region. TFTR!

PR #25, opened 2018-01-25
Tweak README

benesch:
Wrap at 80 chars and add warnings for outside contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,57 +1,73 @@
-# Setup
+## roachprod
 
-Make sure you have gcloud [installed](https://cloud.google.com/sdk/downloads) and configured (`gcloud auth list` to check, `gcloud auth login` to authenticate). You may want to update old installations (`gcloud components update`).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,57 +1,73 @@
...
-# Setting up a cluster using roachprod/roachperf
+## Cluster quick-start using roachprod/roachperf
 
 ```bash
-# Create cluster with 4 nodes and local SSD (last node is used as load generator by roachperf)
+# Create a cluster with 4 nodes and local SSD.
+# The last node is used as load generator by roachperf.
 export NAME="test"
 export FULLNAME="${USER}-${NAME}"
 roachprod create ${NAME} -n 4 --local-ssd
```
petermattis: This should be `roachprod create ${FULLNAME}`. Mea culpa for not updating this documentation when I made that change. Can you fix the use of `${NAME}` here and below?

mberhault:
LGTM after the `${NAME}` fix.

benesch:
Done, thanks!

PR #27, opened 2018-01-26
Prototype of nightly test infra

tschottdorf:
This is an attempt to lay out a lower layer on which to build both
nightly tests (expanded upon in some detail here) and cli-driven
tools (i.e., mostly today's roachperf). This builds, but does not
work (most methods are stubs, but a lot of the required code already
exists in `roachprod/roachperf` and can be lifted).

These will naturally exist on various levels of abstraction. Some tests
may want to mount virtual file systems (to test corruption) and invoke
adversarial I/O workloads on the machines directly. Most performance
tests are much less concerned with these details and just need to be
told a number of Postgres endpoints and machines to run load generators
on.

Please take a look at this example code and provide criticism and
feedback. My impression is that this is a clean, elegant and at the
same time malleable foundation, and every test I have considered
wanting to write (there are a lot) seemed easy to string together
with these primitives.

The goal was to have a clear separation of concerns:

1. A *library* that provisions hardware (`roachprodng`, to be renamed
    when there's less confusion about names. I would probably just call this
    `provision` but there will be an opportunity to bikeshed).
    `roachprodng` does not know anything about CockroachDB.

    In the current code, this library exposes some details about the
    nodes (IP addresses, available mount points) and lets you SSH/scp into
    them, but not much else.

    As a convenience, `roachprodng` allows template interpolation to
    access the provisioned node state, which makes it natural to execute
    commands on the nodes. For example, `--store {{Store 0}}` expands to
    something like `--store /mnt/data0` where `/mnt/data0` could be a
    local SSD (if that's what `roachprodng` was told to set up). This
    avoids leaking provider-specific details into the tests and should
    make it relatively straightforward to target AWS later. (Aside from
    that, it also de-noises the tests and has been pleasant to work with).

2. The `binfetcher` library is used to procure binaries, which includes
`cockroach`, but also `workload`. Binaries can be put anywhere using
`roachprodng`.

3. SQL-specific operations usually rely on `cmd/workload`, which is
    invoked on the nodes to recreate and/or ingest fixtures, and serves as
    the general-purpose SQL runner, so that many tests will simply need a
    cluster and the information on where to invoke `workload`. (It's equally
    straightforward to create a manual SQL connection in a test when that
    is all that's desired: the endpoints are readily available).

4. Various other mechanisms are modular.

    For example, many tests will want to assert the continuous uptime of the
    cluster and can simply run a component that periodically verifies this
    (through SQL). Similarly, other tests may want to periodically run
    `SCRUB` (which can be done through `cmd/workload`).

    Yet another test may want to run chaos and can tap into a modular
    component that fits in easily, though they now have to make sure that
    the load they're running tolerates errors better, and they need to
    make the monitoring component aware of the chaos events.

Note that the nightly benchmark tests are intentionally a bit
simplistic.  We'll likely want to tap into the existing `testRun` struct
from `roachperf`. This is intentional to keep the noise level down; some
of that integration is likely to happen inside of `cmd/workload` anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/27)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,16 +2,167 @@
 
 
 [[projects]]
+  name = "github.com/Shopify/sarama"
+  packages = ["."]
+  revision = "3b1b38866a79f06deddf0487d5c27ba0697ccd65"
+  version = "v1.15.0"
+
+[[projects]]
+  name = "github.com/apache/thrift"
+  packages = ["lib/go/thrift"]
+  revision = "327ebb6c2b6df8bf075da02ef45a2a034e9b79ba"
+  version = "0.11.0"
+
+[[projects]]
+  name = "github.com/certifi/gocertifi"
+  packages = ["."]
+  revision = "deb3ae2ef2610fde3330947281941c562861188b"
+  version = "2018.01.18"
+
+[[projects]]
+  name = "github.com/cockroachdb/cockroach"
+  packages = ["pkg/build","pkg/settings","pkg/util/binfetcher","pkg/util/caller","pkg/util/color","pkg/util/envutil","pkg/util/fileutil","pkg/util/humanizeutil","pkg/util/log","pkg/util/log/logflags","pkg/util/syncutil","pkg/util/timeutil","pkg/util/tracing"]
+  revision = "098b426bd50d40d90a3ff02f9fe7df738629629f"
+  version = "v2.0-alpha.20180122"
+
+[[projects]]
+  name = "github.com/davecgh/go-spew"
+  packages = ["spew"]
+  revision = "346938d642f2ec3594ed81d874461961cd0faa76"
+  version = "v1.1.0"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/dustin/go-humanize"
+  packages = ["."]
+  revision = "bb3d318650d48840a39aa21a027c6630e198e626"
+
+[[projects]]
+  name = "github.com/eapache/go-resiliency"
+  packages = ["breaker"]
+  revision = "6800482f2c813e689c88b7ed3282262385011890"
+  version = "v1.0.0"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/eapache/go-xerial-snappy"
+  packages = ["."]
+  revision = "bb955e01b9346ac19dc29eb16586c90ded99a98c"
+
+[[projects]]
+  name = "github.com/eapache/queue"
+  packages = ["."]
+  revision = "44cc805cf13205b55f69e14bcb69867d1ae92f98"
+  version = "v1.1.0"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/getsentry/raven-go"
+  packages = ["."]
+  revision = "221b2b44fb33f84ed3ea13f3aed62ff48c85636b"
+  source = "https://github.com/cockroachdb/raven-go"
+
+[[projects]]
+  name = "github.com/go-logfmt/logfmt"
+  packages = ["."]
+  revision = "390ab7935ee28ec6b286364bba9b4dd6410cb3d5"
+  version = "v0.3.0"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/gogo/protobuf"
+  packages = ["proto","sortkeys","types"]
+  revision = "26de2f9a7d3b1826f275e7a97c8cad5080589426"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/golang/protobuf"
+  packages = ["proto","protoc-gen-go/descriptor","ptypes","ptypes/any","ptypes/duration","ptypes/timestamp"]
+  revision = "925541529c1fa6821df4e44ce2723319eb2be768"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/golang/snappy"
+  packages = ["."]
+  revision = "553a641470496b2327abcac10b36396bd98e45c9"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/hashicorp/go-version"
+  packages = ["."]
+  revision = "4fe82ae3040f80a03d04d2cccb5606a626b8e1ee"
+
+[[projects]]
   name = "github.com/inconshreveable/mousetrap"
   packages = ["."]
   revision = "76626ae9c91c4f2a10f34cad8ce83ea42c93bb75"
   version = "v1.0"
 
 [[projects]]
+  branch = "master"
+  name = "github.com/kr/logfmt"
+  packages = ["."]
+  revision = "b84e30acd515aadc4b783ad4ff83aff3299bdfe0"
+
+[[projects]]
+  name = "github.com/lightstep/lightstep-tracer-go"
+  packages = [".","collectorpb","lightstep_thrift","lightsteppb","thrift_0_9_2/lib/go/thrift"]
+  revision = "f9f54d495e5c6752b58fae6b66a5cb1107e17b29"
+  version = "v0.15.2"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/opentracing-contrib/go-observer"
+  packages = ["."]
+  revision = "a52f2342449246d5bcc273e65cbdcfa5f7d6c63c"
+
+[[projects]]
+  name = "github.com/opentracing/opentracing-go"
+  packages = [".","ext","log"]
+  revision = "1949ddbfd147afd4d964a9f00b24eb291e0e7c38"
+  version = "v1.0.2"
+
+[[projects]]
+  name = "github.com/openzipkin/zipkin-go-opentracing"
+  packages = [".","flag","thrift/gen-go/scribe","thrift/gen-go/zipkincore","types","wire"]
+  revision = "45e90b00710a4c34a1a7d8a78d90f9b010b0bd4d"
+  version = "v0.3.2"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/petermattis/goid"
+  packages = ["."]
+  revision = "176e84a949d354dcef722e602b312b1a75bacddc"
+
+[[projects]]
+  name = "github.com/pierrec/lz4"
+  packages = ["."]
+  revision = "2fcda4cb7018ce05a25959d2fe08c83e3329f169"
+  version = "v1.1"
+
+[[projects]]
+  name = "github.com/pierrec/xxHash"
+  packages = ["xxHash32"]
+  revision = "f051bb7f1d1aaf1b5a665d74fb6b0217712c69f7"
+  version = "v0.1.1"
+
+[[projects]]
   name = "github.com/pkg/errors"
   packages = ["."]
   revision = "645ef00459ed84a119197bfb8d8205042c6df63d"
   version = "v0.8.0"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/rcrowley/go-metrics"
+  packages = ["."]
+  revision = "8732c616f52954686704c8645fe1a9d59e9df7c1"
+
+[[projects]]
+  name = "github.com/sasha-s/go-deadlock"
+  packages = ["."]
+  revision = "03d40e5dbd5488667a13b3c2600b2f7c2886f02f"
+  version = "v0.2.0"
 
 [[projects]]
   name = "github.com/spf13/cobra"
@@ -28,7 +179,7 @@
 [[projects]]
   branch = "master"
   name = "golang.org/x/net"
-  packages = ["context"]
+  packages = ["context","http2","http2/hpack","idna","internal/timeseries","lex/httplex","trace"]
   revision = "0ed95abb35c445290478a5348a7b38bb154135fd"
 
 [[projects]]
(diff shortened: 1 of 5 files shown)


petermattis:
Do we want a library version of `roachprod`? I suppose it makes some things more convenient. Or perhaps we should make a library version of `roachprod` and have the command line `roachprod` tool make calls into the library. Probably worthwhile to expose the `binfetcher` functionality in `roachprod` as well. See also @danhhz's PR which execs `roachprod` and `roachperf` to run the tests. That can be seen as similar to `roachprod's` use of `gcloud` rather than trying to use some native API. I do like the template interpolation. I wonder if something similar can be achieved on the `roachprod/roachperf` command line. 

I think each nightly test should be its own `Test*()` function. By default they should test against the current source tree instead of downloading `LATEST`. A small script can download `LATEST` if we want to test that, but I'm thinking about development and debugging where you want to test what is in your source tree.

Lastly, the above are small issues (which means we'll likely bikeshed them). In my opinion, monitoring and determination of success pass/fail is the bigger stumbling block. 

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/27#-:-L3oHZrj3giuEJw13PKd:b9s8e3l)*

tschottdorf:
> Or perhaps we should make a library version of roachprod and have the command line roachprod tool make calls into the library. 

That's what's suggested here. Sorry if that wasn't clear. This is mostly adding some sugar to the existing `roachprod`.

> By default they should test against the current source tree instead of downloading LATEST

That's a good point. Once developed and not broken, these tests basically always want to run LATEST (or specific versions such as releases) though, so there's a balance to strike here. I added a commit that I think demonstrates something that should work well: a thin layer of abstraction over `binfetcher` that allows devs to change one thing in the source to prefer a locally built cockroach. Let me know if that seems workable to you (we could consider choosing the binary source based on a flag, etc, too -- but this is fairly discoverable already).

> In my opinion, monitoring and determination of success pass/fail is the bigger stumbling block.

What would you like me to discuss to fill that gap? Monitoring uptime is straightforward (it's in this PR, though I haven't made an attempt at writing the underlying code -- but that is easy, you just periodically connect to SQL and measure uptime. That may not be possible today, but that's easy to fix). Maybe we're talking different kinds of monitoring, but my take-away is that any kind of monitoring that can be done by periodically contacting the nodes in the server is straightforward. For example, we can take automatic heap profiles at fixed intervals during these tests and declare something failed when there's too much allocated (though that's an esoteric one), or we can scrape prometheus and act on that. There sure is a bunch of code we'd have to write if we wanted that, but it seems "straightforward".

Regarding pass/fail, I think that's fairly straightforward as well for large classes of tests (correctness tests: easy, perf tests, easy, "will this cluster recover from this": easy, will disk space be recovered: easy, will this zone config apply within a reasonable amount of time: still easy enough). The only problem I see is that we'll sometimes have to set arbitrary thresholds and those depend on the provisioned hardware (and, for some workloads, there might be a lot of variance).

But even then you can often add a warm-up phase to the test to get stable numbers first before running the actual test (though then what you're really better off doing is maybe making it a pure perf test and just observing the numbers over time). For example, when measuring how rebalancing impacts throughput, you measure throughput first and then decree that during rebalancing, the throughput shouldn't dip below 80% of the original over 10s time windows.

I'm curious to hear what your remaining concerns are.

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/27#-:-L3s0aESbmZJVC1bRtLx:b-lkzejd)*

petermattis:
> That's what's suggested here. Sorry if that wasn't clear. This is mostly adding some sugar to the existing roachprod.

Ah, got it.

> What would you like me to discuss to fill that gap? Monitoring uptime is straightforward (it's in this PR, though I haven't made an attempt at writing the underlying code -- but that is easy, you just periodically connect to SQL and measure uptime. That may not be possible today, but that's easy to fix). Maybe we're talking different kinds of monitoring, but my take-away is that any kind of monitoring that can be done by periodically contacting the nodes in the server is straightforward. For example, we can take automatic heap profiles at fixed intervals during these tests and declare something failed when there's too much allocated (though that's an esoteric one), or we can scrape prometheus and act on that. There sure is a bunch of code we'd have to write if we wanted that, but it seems "straightforward".

I think filling in the gap will require experimentation and prototyping and experience. Some of the monitoring is straightforward. A lot of it seems complex. You mention measuring heap usage. That's a good one where we should be monitoring both Go and Cgo heap usage. How does a test define the normal bounds in such a way that the monitoring is both sensitive enough to catch problems but not so overly sensitive that it fails all the time? Ideally, we would have caught the RocksDB iterator leak through such monitoring. Another example is the recent split issue. Checking for an expected number of ranges seems relatively straightforward. My concern is that we have hundreds of metrics. Are we going to manually add rules and expectations for all of them? 

Re: using prometheus: @danhhz had a similar idea. Setting up prometheus for each test seems a bit onerous. Can we scrape our existing time series metrics and act on those? See [#21833](https://github.com/cockroachdb/cockroach/pull/21833). There is lots of literature on time series anomaly detection and tons of code in various languages. The big challenge is to do something practical and not embark on a boondoggle.

> Regarding pass/fail, I think that's fairly straightforward as well for large classes of tests (correctness tests: easy, perf tests, easy, "will this cluster recover from this": easy, will disk space be recovered: easy, will this zone config apply within a reasonable amount of time: still easy enough). The only problem I see is that we'll sometimes have to set arbitrary thresholds and those depend on the provisioned hardware (and, for some workloads, there might be a lot of variance).

I like your optimism regarding the ease. Chaos tests seem a bit challenging on the surface, though really we should do some experimentation with something straightforward and see whether it works or not. 

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[infra_helpers.go, line 97 at r2](https://reviewable.io:443/reviews/cockroachdb/roachprod/27#-L3sNi551rgtzJGp97Gl:-L3sNi551rgtzJGp97Gm:b-f2fclm) ([raw file](https://github.com/cockroachdb/roachprod/blob/11410e22382d278a891c06841f8b47277ba47dd2/infra_helpers.go#L97)):*
> ```Go
> 			return nil, errors.Wrapf(err, "stat %s", localBin)
> 		}
> 		log.Printf("%s not found in local repo; downloading", binName)
> ```

I would think it should be an error if the user asked for a local binary and we couldn't find one. They probably just forgot to build it.

---

*[infra_test.go, line 42 at r2](https://reviewable.io:443/reviews/cockroachdb/roachprod/27#-L3sNuOgnbER96AKjOUm:-L3sNuOgnbER96AKjOUn:b-thoo8y) ([raw file](https://github.com/cockroachdb/roachprod/blob/11410e22382d278a891c06841f8b47277ba47dd2/infra_test.go#L42)):*
> ```Go
> 	}
> 
> 	// NB: change this to LocalDevelopmentSource during development, so that
> ```

My guess is that this should be configurable on the command line. Also, we probably want to make `LATEST` configurable so that we can test older versions and patch releases. Can probably combine the two by looking for `local` (or `LOCAL`) as a version.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/27#-:-L3sOKVW9WUlrqxjugxH:b7ajufx)*

tschottdorf:
> I think filling in the gap will require experimentation and prototyping and experience.

Agreed, but that's something that will (have to) happen over time (and until then, specific criteria in specific tests is already a big win and easy to do) and doesn't really impact what's presented here much (or do you think it does?).

> There is lots of literature on time series anomaly detection and tons of code in various languages. The big challenge is to do something practical and not embark on a boondoggle.

What I like about performance-thresholded pass/fail criteria is that they'll force us to make it easy to obtain some of these numbers programmatically. Perhaps the right thing is to expose some time series aggregation into our SQL dialect, which on the face of it doesn't seem so hard (after all, it's already in the UI graphs). To get industry-grade anomaly detection (I hope such a thing exists), we'll certainly have to build some auxiliary infra and feed into that.
Similar to the above point, I don't think we need to figure that out 100% right now. We might be doing simplistic things at first, but node statement statistics, loadgen-reported QPS/latencies, and periodic polling of the prometheus endpoint, etc already get us somewhere, and it doesn't bar us from getting fancier over time.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[infra_helpers.go, line 97 at r2](https://reviewable.io:443/reviews/cockroachdb/roachprod/27#-L3sNi551rgtzJGp97Gl:-L3tiKmUDbK1sF4QrBSX:b-na5ocb) ([raw file](https://github.com/cockroachdb/roachprod/blob/11410e22382d278a891c06841f8b47277ba47dd2/infra_helpers.go#L97)):*

I would think it should be an error if the user asked for a local binary and we couldn't find one. They probably just forgot to build it.
</blockquote></details>

Yeah, this should take a slice of binary names in the constructor and then require that those are found in the "search path" (where the "search path" is probably the repo root and $GOPATH/bin).

---

*[infra_test.go, line 42 at r2](https://reviewable.io:443/reviews/cockroachdb/roachprod/27#-L3sNuOgnbER96AKjOUm:-L3tXap_ckA3MfVzJm4A:bxoz10s) ([raw file](https://github.com/cockroachdb/roachprod/blob/11410e22382d278a891c06841f8b47277ba47dd2/infra_test.go#L42)):*

My guess is that this should be configurable on the command line. Also, we probably want to make `LATEST` configurable so that we can test older versions and patch releases. Can probably combine the two by looking for `local` (or `LOCAL`) as a version.
</blockquote></details>

I think what makes sense is to aggressively factor out tests that are of the "just give me any cluster and I'll do my thing" variety and put them into `cmd/workload` where possible. Then, make a subpackage (say `workloadtest`) whose job it is to offer runs of any workload with provisioning and teardown of a corresponding cluster (that package can also absorb tests that need more than `cmd/workload` but can also run against any cluster at any recent enough version).

Then, flags make a lot of sense on that package without imposing themselves on other, more opinionated tests. As a bonus, anything in that package can be offered via `roachprod-cli` to be run against a user-provisioned cluster.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/27#-:-L3ti_zAMa61TvGFnGFO:b-lk76zj)*

petermattis:
> Agreed, but that's something that will (have to) happen over time (and until then, specific criteria in specific tests is already a big win and easy to do) and doesn't really impact what's presented here much (or do you think it does?).

My only concern is that we lull ourselves into thinking that what is presented here is sufficient. I think between this and work @danhhz there is a good start, but there is plenty of additional experimentation and enhancement in our near future. 

> What I like about performance-thresholded pass/fail criteria is that they'll force us to make it easy to obtain some of these numbers programmatically. Perhaps the right thing is to expose some time series aggregation into our SQL dialect, which on the face of it doesn't seem so hard (after all, it's already in the UI graphs). 

Is there a reason the time series needs to be exposed via SQL? Tests can access the timeseries via a grpc endpoint (`ts/tspb.TimeSeries.Query`). I'd expect any anomaly detection based pass/fail to be accessed programmatically, either by linking in a library of calling out to an external tool.

> To get industry-grade anomaly detection (I hope such a thing exists), we'll certainly have to build some auxiliary infra and feed into that. Similar to the above point, I don't think we need to figure that out 100% right now. We might be doing simplistic things at first, but node statement statistics, loadgen-reported QPS/latencies, and periodic polling of the prometheus endpoint, etc already get us somewhere, and it doesn't bar us from getting fancier over time.

Agreed we don't need to figure this out 100% right now.

What is your plan for this PR given your imminent absence?

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[infra_helpers.go, line 97 at r2](https://beta.reviewable.io/reviews/cockroachdb/roachprod/27#-L3sNi551rgtzJGp97Gl:-L3yJBAZ0vbl1jsXdk6g:b-o1ob3l) ([raw file](https://github.com/cockroachdb/roachprod/blob/11410e22382d278a891c06841f8b47277ba47dd2/infra_helpers.go#L97)):*
<details><summary><i>Previously, tschottdorf (Tobias Schottdorf) wrote…</i></summary><blockquote>

Yeah, this should take a slice of binary names in the constructor and then require that those are found in the "search path" (where the "search path" is probably the repo root and $GOPATH/bin).
</blockquote></details>

For those of us running Mac OS X, `$GOPATH/bin`is not going to contain binaries that are runnable on linux. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/27#-:-L3yJap8Er3FlpOaKTY3:b4kn329)*

tschottdorf:
> Is there a reason the time series needs to be exposed via SQL?

Not for our purposes, but generally it would be good to make time series available from within SQL generally. It'd also streamline things a bit, but there's no reason we *have* to do it.

> What is your plan for this PR given your imminent absence?

Hope that some of it will be found useful by whoever will actually work on this.

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/27#-:-L40OjzI179YcEJof6gA:b-ip2b0j)*

PR #28, opened 2018-01-26
Vendor roachperf deps

benesch:
Forgot to include these in the roachperf merge (5ff821aec).

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/28)


benesch:
Urgh, the AWS library is huge. Are we ok with committing all this directly?

petermattis:
Where is the aws dependency coming from? Ah, `upload`. Yeah, I'm fine with this.

PR #33, opened 2018-01-29
Rename "monitor" command to "gc"

petermattis:
This is in anticipation of adding a new "monitor" command which monitors
a single cluster.

Add a few newlines to the short command help in order to group the
commands into sections.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/33)


petermattis:
@mberhault Where is the cronjob that runs this periodically? Should we have the periodic running of `roachprod gc` be performed by teamcity? Small challenge is to figure out what to do with the `roachprod.tracking.txt` file.

mberhault:
LGTM. It's part of the usual cron jobs on the shared machine: https://github.com/cockroachlabs/production/tree/master/shared. You'll need to push the binary and change the wrapper script in the binaries subdirectory.

PR #44, opened 2018-02-02
Wait for nodes to start after creating a cluster

petermattis:
This makes it easier to script usage of roachprod as the user no longer
has to insert a sleep or other mechanism to determine when the nodes in
the newly created cluster are up.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/44)

### DIFF
--- a/cluster_synced.go
+++ b/cluster_synced.go
@@ -286,6 +286,37 @@ func (c *syncedCluster) run(w io.Writer, nodes []int, title, cmd string) error {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


petermattis:
> Might be worth adding a timeout to wait() so that we eventually give up if a node fails to come up.

Yeah, maybe. I'll add a TODO. If we see problems in nightlies we can address it, though I'm hesitant to preemptively do anything.

PR #45, opened 2018-02-02
Don't display spinners when stdout is not a terminal

petermattis:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/45)


petermattis:
The code to do this is a bit messy, but I don't want to invest energy in a more holistic approach at this time.

writer.go
```diff
@@ -4,9 +4,14 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
 	"strings"
+
+	isatty "github.com/mattn/go-isatty"
```
benesch: You could avoid this dep with https://godoc.org/golang.org/x/crypto/ssh/terminal#IsTerminal, I think.

petermattis: Ok, let me try that.

petermattis:
> In the future, could you separate out the dep changes into a separate commit, plz? It was hard to navigate the review, even when using github instead of reviewable and with most of them unexpanded

Yeah, I realize the dep changes obfuscated the actual code changes. If the dep changes are in a 2nd commit, then the first commit won't build. How do you put the dep changes in a first commit when the dependency doesn't exist yet?

writer.go
petermattis: Ok, that avoided the dependency on `mattn/go-isatty`, but `dep` still needed to pull in a bunch of new dependencies.

petermattis:
@danhhz Ok, I'll try both your suggestions next time this comes up.

PR #46, opened 2018-02-05
Learn to recursively get directories

benesch:
Teach `roachprod get` to handle directories properly.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/46)

### DIFF
--- a/cluster_synced.go
+++ b/cluster_synced.go
@@ -524,7 +524,8 @@ func (c *syncedCluster) get(src, dest string) {
 				defer session.Close()
 				dest := dest
 				if len(c.nodes) > 1 {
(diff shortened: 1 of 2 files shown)


benesch:
FWIW it seems a little crazy to me that we're growing our own SCP implementation, but ¯\_(ツ)_/¯

petermattis:
> FWIW it seems a little crazy to me that we're growing our own SCP implementation, but ¯_(ツ)_/¯

Yeah, as we add functionality we need to keep an eye on this. I wouldn't mind switching to exec'ing `scp` for both `get` and `put`. We'd lose progress indicators, but we could replace them with spinners.

benesch:
:tada: thanks for the review!

PR #48, opened 2018-02-05
Optionally start nodes sequentially

### LINKED ISSUE #47: feature request: Align node IDs with instance names
It always annoys me when I'm looking into an issue that node IDs never correspond with the number in the node address/name. I end having to keep a note handy so I can mentally translate as needed.

Would others be ok with modifying the startup process slightly to avoid this? For example, ditching the use of `cockroach init`? Or starting the cluster with just one node and running `cockroach init`, then adding the nodes in order after that.

Or am I the only one really bothered by this?

benesch:
Provide a flag that starts nodes sequentially so that node IDs match
hostnames.

Fix #47.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/48)

### DIFF
--- a/cockroach.go
+++ b/cockroach.go
@@ -5,13 +5,22 @@ import (
 	"strings"
 )
 
+var startOpts struct {
+	sequential bool
+}
+
 type cockroach struct{}
 
 func (r cockroach) start(c *syncedCluster) {
 	display := fmt.Sprintf("%s: starting", c.name)
 	host1 := c.host(1)
 	nodes := c.serverNodes()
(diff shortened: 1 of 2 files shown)


petermattis:
:lgtm: that was easy.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/48#-:-L4ca-qVnaHaWZi4n3RO:b827n1n)*

PR #58, opened 2018-02-09
Avoid using gcloud when manipulating local clusters

petermattis:
This speeds up local cluster creation and destruction by about an order
of magnitude due to not having to list the cloud clusters.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/58)


petermattis:
Added a second commit which fixes the problem I was seeing with `roachprod destroy local` sometimes failing.

PR #65, opened 2018-02-17
Rework how local clusters find binaries

petermattis:
Previously, local clusters were only able to start binaries using an
absolute path. The `put` command was also strange in that it violated
the appearance of multiple nodes as each node would target the same
directory. Now, local nodes are each given their own directory and
relative paths for specifying the binary to run
work (e.g. `./cockroach`).

See https://github.com/cockroachdb/cockroach/issues/22640

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/65)


petermattis:
The implementation is a bit diffuse, but we already have a lot of local cluster special casing. On the flip side, this allows removing special casing for local clusters for `roachprod` users. See https://github.com/cockroachdb/cockroach/pull/22791.

PR #67, opened 2018-02-19
Make jsonVM type solely an implementation detail of gcloud.

bobvawter:
The jsonVM type is specific to the JSON format returned from gcloud invocations
and doesn't make sense in the context of other cloud providers.  This change
replaces the use of jsonVM in the Cloud type with the provider-agnostic VM
type.

Other changes:
* The VM type is moved to its own file.
* jsonVM -> VM conversion code moved to a toVM() function.
* Cloud instance error reporting is generalized.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/67)

### DIFF
--- a/cloud/cluster_cloud.go
+++ b/cloud/cluster_cloud.go
@@ -3,10 +3,9 @@ package cloud
 import (
 	"bytes"
 	"fmt"
-	"log"
 	"os"
 	"path/filepath"
-	"regexp"
+	"sort"
 	"strings"
 	"sync"
 	"text/tabwriter"
@@ -16,22 +15,35 @@ import (
 	"github.com/pkg/errors"
 )
 
-const vmNameFormat = "user-<clusterid>-<nodeid>"
-
 type Cloud struct {
 	Clusters map[string]*CloudCluster
-	// Individual "bad" instances by category.
-	InvalidName  JsonVMList
-	NoExpiration JsonVMList
-	BadNetwork   JsonVMList
+	// Any VM in this list can be expected to have at least one element
+	// in its Errors field.
+	BadInstances VMList
+}
+
+// Collate Cloud.BadInstances by errors.
+func (c *Cloud) BadInstanceErrors() map[error]VMList {
+	ret := map[error]VMList{}
+
+	// Expand instances and errors
+	for _, vm := range c.BadInstances {
+		for _, err := range vm.Errors {
+			ret[err] = append(ret[err], vm)
+		}
+	}
+
+	// Sort each VMList to make the output prettier
+	for _, v := range ret {
+		sort.Sort(v)
+	}
+
+	return ret
 }
 
 func newCloud() *Cloud {
 	return &Cloud{
-		Clusters:     make(map[string]*CloudCluster),
-		InvalidName:  make(JsonVMList, 0),
-		NoExpiration: make(JsonVMList, 0),
-		BadNetwork:   make(JsonVMList, 0),
+		Clusters: make(map[string]*CloudCluster),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


petermattis:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[main.go, line 371 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/67#-L5ix6I3Hn2hHnDtNVk2:-L5ix6I3Hn2hHnDtNVk3:by2r9ls) ([raw file](https://github.com/cockroachdb/roachprod/blob/e1274f3c94e46423518f8ee8a1eeb673c4ea7a0a/main.go#L371)):*
> ```Go
> 		// Optionally print any dangling instances with errors
> 		if listDetails {
> 			for err, vmList := range cloud.BadInstanceErrors() {
> ```

The iteration order over maps is randomized which will cause the errors to be output in random order. Not sure if this is worth fixing or not. The standard way to fix is to extract the keys (i.e. `err`) into a list, sort and then perform a second loop over the sorted errors.

---

*[cloud/cluster_cloud.go, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/67#-L5ixTCqJX9fCRiwgtJo:-L5ixTCqJX9fCRiwgtJp:b-vmewo5) ([raw file](https://github.com/cockroachdb/roachprod/blob/e1274f3c94e46423518f8ee8a1eeb673c4ea7a0a/cloud/cluster_cloud.go#L34)):*
> ```Go
> 			if _, ok := ret[err]; !ok {
> 				ret[err] = make(VMList, 0)
> 			}
> ```

This block isn't necessary as the zero value for a slice can be appended to.

---

*[cloud/cluster_cloud.go, line 50 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/67#-L5ix_5wcQcaBbWCCZ_U:-L5ix_5wcQcaBbWCCZ_V:bxbhszn) ([raw file](https://github.com/cockroachdb/roachprod/blob/e1274f3c94e46423518f8ee8a1eeb673c4ea7a0a/cloud/cluster_cloud.go#L50)):*
> ```Go
> 	return &Cloud{
> 		Clusters:     make(map[string]*CloudCluster),
> 		BadInstances: make(VMList, 0),
> ```

Similar to another comment, this isn't necessary (and neither was the previous code): the zero value for a slice can be appended to.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/67#-:-L5ixmdSmAUQfneuTk_H:bnfp4nl)*

bobvawter:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[main.go, line 371 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/67#-L5ix6I3Hn2hHnDtNVk2:-L5nvJifREYe7Ji3pFOb:bostppk) ([raw file](https://github.com/cockroachdb/roachprod/blob/e1274f3c94e46423518f8ee8a1eeb673c4ea7a0a/main.go#L371)):*

The iteration order over maps is randomized which will cause the errors to be output in random order. Not sure if this is worth fixing or not. The standard way to fix is to extract the keys (i.e. `err`) into a list, sort and then perform a second loop over the sorted errors.
</blockquote></details>

Added a collation step to keep the error messages in stable order.

---

*[cloud/cluster_cloud.go, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/67#-L5ixTCqJX9fCRiwgtJo:-L5nvTYTMST4TVO8W3pU:bb6un1j) ([raw file](https://github.com/cockroachdb/roachprod/blob/e1274f3c94e46423518f8ee8a1eeb673c4ea7a0a/cloud/cluster_cloud.go#L34)):*

This block isn't necessary as the zero value for a slice can be appended to.
</blockquote></details>

Done

---

*[cloud/cluster_cloud.go, line 50 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/67#-L5ix_5wcQcaBbWCCZ_U:-L5nvUD0CRB4y2xEKyV7:bb6un1j) ([raw file](https://github.com/cockroachdb/roachprod/blob/e1274f3c94e46423518f8ee8a1eeb673c4ea7a0a/cloud/cluster_cloud.go#L50)):*

Similar to another comment, this isn't necessary (and neither was the previous code): the zero value for a slice can be appended to.
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/67)*

petermattis:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[ui/collate_errors.go, line 14 at r2](https://reviewable.io:443/reviews/cockroachdb/roachprod/67#-L5oe2mHgJVuXkM2cF10:-L5oe2mHgJVuXkM2cF11:b-ci05qw) ([raw file](https://github.com/cockroachdb/roachprod/blob/9f9931fc1e6c66ef8914609f8d7ca40f0f663e50/ui/collate_errors.go#L14)):*
> ```Go
> func (l ErrorsByError) Swap(i, j int) {
> 	l[i], l[j] = l[j], l[i]
> }
> ```

FYI, this is the old way of sorting a slice, and while slightly faster, isn't necessary for this code. Go1.9 (or perhaps go1.8) introduced `sort.Slice`:

```
  var errors []error
  ...
  sort.Slice(errors, func(i, j int) bool {
    return errors[i].Error() < errors[j].Error()
  })
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/67)*

PR #69, opened 2018-02-20
add sql subcommand

benesch:
Add a `roachprod sql` subcommand that replicates the behavior of
`crl-sql`. When no arguments are provided, it opens a SQL shell on the
specified node. When arguments are provided, it executes `cockroach sql
ARGS` on all specified nodes.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/69)


benesch:
Reviewed as part of #70.

PR #72, opened 2018-02-20
Don't crash if SSH keys don't exist

benesch:
When a default SSH key doesn't exist, rather than returning a nil
signer, which causes a segfault inside of the ssh package, avoid
appending the signer at all.

Also return a nicer error message when keys are encrypted.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/72)

### DIFF
--- a/ssh/ssh.go
+++ b/ssh/ssh.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/72#-:-L5ocLK8d29XvC0htOL7:bnfp4nl)*

danhhz:
it's no longer crashing for me, but i seem to get the `skipping encrypted SSH key "/Users/dan/.ssh/id_rsa"; if necessary, add the key to your SSH agent` message even after i've added it to my agent

benesch:
Yeah, that's expected. The code isn't smart enough to detect if you've added the key to your agent. I'm going to merge because this is strictly better than before, but ping me if it drives you nuts and I can whip something up.

danhhz:
This prints out several times per test, so it's pretty spammy right now

PR #73, opened 2018-02-20
install: anchor pkill regex

### LINKED ISSUE #60: `roachprod stop` kills too much
If running a local cluster, `roachprod stop cockroach` will terminate a running Cockroach build, not just the nodes in the cluster. Perhaps `killall` would work better than `pkill`?

benesch:
"kv" is a rather short string; let's make sure it matches exactly.

Possibly fix #60.

---

I can no longer repro #60 (shrug), but this seems worth doing anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/73)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -112,7 +112,7 @@ func (c *SyncedCluster) Stop() {
 		}
 		defer session.Close()
 
-		cmd := `pkill -9 "cockroach|java|mongo|kv|ycsb" || true ;
+		cmd := `pkill -9 "^(cockroach|java|mongo|kv|ycsb)$" || true ;
 `
 		cmd += fmt.Sprintf("kill -9 $(lsof -t -i :%d -i :%d) 2>/dev/null || true ;\n",
 			Cockroach{}.NodePort(c, c.Nodes[i]),


install/cluster_synced.go
```diff
@@ -112,7 +112,7 @@ func (c *SyncedCluster) Stop() {
 		}
 		defer session.Close()
 
-		cmd := `pkill -9 "cockroach|java|mongo|kv|ycsb" || true ;
+		cmd := `pkill -9 "^(cockroach|java|mongo|kv|ycsb)$" || true ;
```
petermattis: Killing `java` probably isn't a good idea on localhost. I bet we can remove `cockroach`, `java` and `mongo` from this list and rely on the `lsof`-based killing below.

tschottdorf: I was thinking the same, killing random processes is a bad idea in general when the `local` cluster is used.

petermattis: Btw, this originated from using `roachperf` on non-ephemeral clusters with cockroach started by other means than `roachperf`. It is very likely overkill.

benesch:
Superseded by #74.

PR #88, opened 2018-02-23
Add --secure flag to adminurl command

petermattis:
Reorganize flags initialization. Some commands were being given flags
(`get --secure`) they didn't use and some commands were not being given
flags they could use (`test --sequential`).

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/88)


petermattis:
> I'm not convinced the real motivation here wasn't making an opportunity to wrap some long lines. :)

Ha!

PR #90, opened 2018-02-24
Accurately display when clusters will be destroyed

petermattis:
Take the frequency of the `roachprod gc` cronjob into account when
computing the lifetime remaining for clusters. Remove the
`--destroy-after` flag which added a confusing extra bit of life to
clusters and had me perpetually confused about when a cluster would
actually be destroyed.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/90)

### DIFF
--- a/cloud/cluster_cloud.go
+++ b/cloud/cluster_cloud.go
@@ -78,8 +78,15 @@ func (c *CloudCluster) ExpiresAt() time.Time {
 	return c.CreatedAt.Add(c.Lifetime)
 }
 
+func (c *CloudCluster) GCAt() time.Time {
+	// NB: GC is performed every hour. We calculate the lifetime of the cluster
+	// taking the GC time into account to accurately reflect when the cluster
+	// will be destroyed.
+	return c.ExpiresAt().Add(time.Hour - 1).Truncate(time.Hour)
+}
+
 func (c *CloudCluster) LifetimeRemaining() time.Duration {
-	return time.Until(c.ExpiresAt())
+	return time.Until(c.GCAt())
 }
 
 func (c *CloudCluster) String() string {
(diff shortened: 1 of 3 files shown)


petermattis:
This will require pushing the new `roachprod` binary to the `shared-0001.crdb.io`. Also, this was a continuous source of confusion for me, but clearly Marc added it for some reason. Perhaps it wasn't confusing to him or to others.

benesch:
> Also, this was a continuous source of confusion for me, but clearly Marc added it for some reason. Perhaps it wasn't confusing to him or to others.

Heh, I'm confused about your confusion. What's the big deal if your cluster lives a little longer than it says it will on the tin? If it's past its expiration you're living on the edge, but at least you have a chance at saving a cluster you forgot about.

Code LGTM, and approach LGTM provided no one who was relying on the old behavior holds me accountable for their missing cluster. :)

petermattis:
> Heh, I'm confused about your confusion. What's the big deal if your cluster lives a little longer than it says it will on the tin? If it's past its expiration you're living on the edge, but at least you have a chance at saving a cluster you forgot about.

My confusion is due to the inaccuracy of the "lives a little longer". I never remember if that little longer is 2 hours or 3 hours or 6 hours. And there is also a small bit of extra time due to `gc` only being run on the hour, so a cluster could live up to 6h59m past its lifetime. I don't see what this extra time is buying other than confusion. If we want clusters to have an 18h lifetime by default, let's just give them that lifetime.

danhhz:
behavior change lgtm. i've been scarred by our leaked terraform clusters and when i see `-4h left`, i start to worry if the auto-shutdown mechanism is broken

petermattis:
> i've been scarred by our leaked terraform clusters and when i see -4h left, i start to worry if the auto-shutdown mechanism is broken

That's exactly how I feel.

petermattis:
This is ready for another look. Similar to before, but clusters are considered "expiring" if they will be destroyed in the next 2 hours.

PR #92, opened 2018-02-26
Remove display of the active account from "list"

petermattis:
This was present from the initial version of roachprod, but it isn't
clear what problem it is solving.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/92)

### DIFF
--- a/main.go
+++ b/main.go
@@ -384,7 +384,6 @@ The default output shows one line per cluster, including the local cluster if
 it exists:
 
   ~ roachprod list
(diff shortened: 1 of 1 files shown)


benesch:
Heh, I didn't realize this was the only difference between `list` and `sync`. I thought `sync` was somehow duplicating effort but now I realize it's just slow.

petermattis:
Ping.

PR #96, opened 2018-02-27
Add AWS support to roachprod

bobvawter:
This adds support for managing clusters in AWS and for creating AWS+GCE
cockroach clusters.  The basic approach for AWS is similar to GCE, however
there are some complications around needing to upload a SSH key to AWS before
creating instances.  Also, AWS requires quite a bit more configuration than
GCE; all of these parameters are managed as flags, should a user need to
override the defaults.

Misc changes:
* Add vm.VM.RemoteUser since the AWS system image uses "ubuntu"
* Add --advertise-host=vm.PublicIP to cockroach clusters running in a
  multi-placement configuration.

To create a multi-cloud cluster:

```
./roachprod create bob-xyzzy -c aws,gce -n 8
```

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/96)

### DIFF
--- a/hosts.go
+++ b/hosts.go
@@ -39,12 +39,10 @@ func syncHosts(cloud *cloud.Cloud) error {
 
 		// Align columns left and separate with at least two spaces.
 		tw := tabwriter.NewWriter(file, 0, 8, 2, ' ', 0)
-		tw.Write([]byte("# user@host\tlocality\n"))
+		tw.Write([]byte("# user@host\tlocality\tvpcId\n"))
 		for _, vm := range c.VMs {
-			// N.B. gcloud uses the local username to log into instances rather
-			// than the username on the authenticated Google account.
 			tw.Write([]byte(fmt.Sprintf(
-				"%s@%s\t%s\n", config.OSUser.Username, vm.PublicIP, vm.Locality())))
+				"%s@%s\t%s\t%s\n", vm.RemoteUser, vm.PublicIP, vm.Locality(), vm.VPC)))
 		}
 		if err := tw.Flush(); err != nil {
 			return errors.Wrapf(err, "problem writing file %s", filename)
@@ -78,7 +76,7 @@ func gcHostsFiles(cloud *cloud.Cloud) error {
 }
 
 func newInvalidHostsLineErr(line string) error {
-	return fmt.Errorf("invalid hosts line, expected <username>@<host> [locality], got %q", line)
+	return fmt.Errorf("invalid hosts line, expected <username>@<host> [locality] [vpcId], got %q", line)
 }
 
 func loadClusters() error {
@@ -105,17 +103,19 @@ func loadClusters() error {
 		}
 
 		for _, l := range lines {
+			// We'll consume the fields as we go along
 			fields := strings.Fields(l)
 			if len(fields) == 0 {
 				continue
 			} else if len(fields[0]) > 0 && fields[0][0] == '#' {
 				// Comment line.
 				continue
-			} else if len(fields) > 2 {
+			} else if len(fields) > 3 {
 				return newInvalidHostsLineErr(l)
 			}
 
 			parts := strings.Split(fields[0], "@")
+			fields = fields[1:]
 			var n, u string
 			if len(parts) == 1 {
 				u = config.OSUser.Username
@@ -127,14 +127,26 @@ func loadClusters() error {
 				return newInvalidHostsLineErr(l)
 			}
 
-			var l string
-			if len(fields) == 2 {
-				l = fields[1]
+			var locality string
+			if len(fields) > 0 {
+				locality = fields[0]
+				fields = fields[1:]
+			}
+
+			var vpc string
+			if len(fields) > 0 {
+				vpc = fields[0]
+				fields = fields[1:]
+			}
+
+			if len(fields) > 0 {
+				return newInvalidHostsLineErr(l)
 			}
 
 			c.VMs = append(c.VMs, n)
 			c.Users = append(c.Users, u)
-			c.Localities = append(c.Localities, l)
+			c.Localities = append(c.Localities, locality)
+			c.VPCs = append(c.VPCs, vpc)
 		}
 		install.Clusters[file.Name()] = c
 	}
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -41,6 +41,7 @@ type SyncedCluster struct {
 	VMs        []string
 	Users      []string
 	Localities []string
+	VPCs       []string
 	// all other fields are populated in newCluster.
 	Nodes   []int
 	LoadGen int
--- a/install/cockroach.go
+++ b/install/cockroach.go
@@ -235,6 +235,17 @@ tar cvf certs.tar certs
 	host1 := c.host(1)
 	nodes := c.ServerNodes()
 
+	// If we're creating nodes that span VPC (e.g. AWS multi-region or
+	// multi-cloud), we'll tell the nodes to advertise their public IPs
+	// so that attaching nodes to the cluster Just Works.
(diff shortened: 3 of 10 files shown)


petermattis:
:lgtm:

Very excited to see this. Might be some buglets, but since you have to opt in to using aws, it shouldn't affect existing users of `roachprod`.

I didn't see mention of the `--local-ssd` flag when creating AWS instances. Did I miss it? Is that still to do?

Should `roachprod list` be extended to indicate which providers a cluster has nodes on? Or is it already possible to glean that info from `roachprod list`. This can be done in a follow-on PR.

---

Review status: 0 of 9 files reviewed at latest revision, 4 unresolved discussions.

---

*[install/cockroach.go, line 243 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/96#-L6NowKx4jEY6D-dxId1:-L6NowKx4jEY6D-dxId2:b-vyntd8) ([raw file](https://github.com/cockroachdb/roachprod/blob/6d12cf1080e8ede8acff8ce7273fd13da28135e4/install/cockroach.go#L243)):*
> ```Go
> 	for i, loc := range c.Localities {
> 		if i > 0 && loc != c.Localities[0] {
> 			advertisePublicIP = true
> ```

Won't this also fire for geo-distributed clusters (i.e. created with the `--geo` flag)?

---

*[vm/aws/aws.go, line 44 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/96#-L6NxJXaSBeTdRkoF9q5:-L6NxJXaSBeTdRkoF9q6:bvb9x3z) ([raw file](https://github.com/cockroachdb/roachprod/blob/6d12cf1080e8ede8acff8ce7273fd13da28135e4/vm/aws/aws.go#L44)):*
> ```Go
> 		[]string{
> 			"us-east-2:ami-965e6bf3",
> 			"us-west-2:ami-79873901",
> ```

How do you locate these AMIs?

---

*[vm/aws/aws.go, line 89 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/96#-L6Ny0yI3neJGEcj-CQA:-L6Ny0yI3neJGEcj-CQB:blyjt83) ([raw file](https://github.com/cockroachdb/roachprod/blob/6d12cf1080e8ede8acff8ce7273fd13da28135e4/vm/aws/aws.go#L89)):*
> ```Go
> // ConfigSSH ensures that for each region we're operating in, we have a <user>-<hash> keypair,
> // where <hash> is a hash of the public key.  We use a hash since a user
> // probably has multiple machines they're running roachprod on and these machines (ought to) have separate
> ```

Nit: we try to limit line length to 100 columns, and often wrap comments at 78 columns (though that isn't always enforced).

---

*[vm/aws/aws.go, line 244 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/96#-L6NyXCE_SoRn3hcXxRk:-L6NyXCE_SoRn3hcXxRl:b5pycx5) ([raw file](https://github.com/cockroachdb/roachprod/blob/6d12cf1080e8ede8acff8ce7273fd13da28135e4/vm/aws/aws.go#L244)):*
> ```Go
> 	var ret vm.List
> 	for _, region := range regions {
> 		vms, err := p.listRegion(region)
> ```

You have to list each region individually? That's irritating.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/96#-:-L6Nz8sBSu-Mr0IA5w_w:b-t0awzy)*

bobvawter:
PTAL

Filed https://github.com/cockroachdb/roachprod/issues/103 for roachprod list

I missed the `--local-ssd` flag.  Looks like that's going to need some additional plumbing to switch up the machine types, since the m5-class only supports EBS volumes.  Filed https://github.com/cockroachdb/roachprod/issues/104 which I'll take care of later today.

---

Review status: 0 of 10 files reviewed at latest revision, 4 unresolved discussions.

---

*[install/cockroach.go, line 243 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/96#-L6NowKx4jEY6D-dxId1:-L6Rf4iCLRRtrlWudWUi:bq0wg0c) ([raw file](https://github.com/cockroachdb/roachprod/blob/6d12cf1080e8ede8acff8ce7273fd13da28135e4/install/cockroach.go#L243)):*

Won't this also fire for geo-distributed clusters (i.e. created with the `--geo` flag)?
</blockquote></details>

Per discussion on Slack, I've plumbed a concept of a VPC into SyncedCluster, since we only need public IPs when going cross-VPC.

---

*[vm/aws/aws.go, line 44 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/96#-L6NxJXaSBeTdRkoF9q5:-L6SCYu9D4RGZ72DlNjN:b-z3l4kr) ([raw file](https://github.com/cockroachdb/roachprod/blob/6d12cf1080e8ede8acff8ce7273fd13da28135e4/vm/aws/aws.go#L44)):*

How do you locate these AMIs?
</blockquote></details>

Added comment with a pointer to Ubuntu's AMI locator

---

*[vm/aws/aws.go, line 89 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/96#-L6Ny0yI3neJGEcj-CQA:-L6SEjhCBUQJgh1mEUlh:b-csclww) ([raw file](https://github.com/cockroachdb/roachprod/blob/6d12cf1080e8ede8acff8ce7273fd13da28135e4/vm/aws/aws.go#L89)):*

Nit: we try to limit line length to 100 columns, and often wrap comments at 78 columns (though that isn't always enforced).
</blockquote></details>

Cleaned up line-length in this file and as I touch other things.

---

*[vm/aws/aws.go, line 244 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/96#-L6NyXCE_SoRn3hcXxRk:-L6SEtQ93i0NkPruMjx3:bzet3w2) ([raw file](https://github.com/cockroachdb/roachprod/blob/6d12cf1080e8ede8acff8ce7273fd13da28135e4/vm/aws/aws.go#L244)):*

You have to list each region individually? That's irritating.
</blockquote></details>

Pretty much every interaction with AWS is region-specific.  I've tweaked this to query in parallel.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/96#-:-L6ReB9O-yc3AZEG52R0:bn9ij98)*

bobvawter:
@benesch, @danhhz:  This changes the locality strings to look like `cloud=aws,region=us-east-1,zone=us-east-1b`.  This makes the drill-down in the admin UI make more sense when running multi-cluster.  Is this going to blow up any expectations elsewhere about what the locality string should look like?

danhhz:
Nothing I can think of for `cloud`, but what do the aws regions/zones look like? The roachmart test currently hardcodes the gce zones with a TODO to figure out how to keep them in sync with roachprod. If the aws zones don't match the gce ones, then I think roachmart wouldn't work on aws. https://github.com/cockroachdb/cockroach/blob/master/pkg/ccl/workloadccl/roachmartccl/roachmart.go#-L26L30

benesch:
Yeah, they won't match match. GCE is `us-east1-b`, while AWS is `us-east-1b`.

I think we should just teach roachmart to introspect the available regions from the `crdb_internal.nodes_gossip` cluster. In the meantime landing this won't break the test on GCE, so I say carry on.

danhhz:
Oh yes, I agree. I should have been explicit that I don't think fixing that TODO should block this PR

petermattis:
:shipit: 

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/96#-:-L6U8veWwiIMUXvTqI3Y:bizlbj6)*

PR #97, opened 2018-02-27
Support filtering clusters in list command

### LINKED ISSUE #93: "list" should take an optional regex filter
The `list` command should take an optional regex filter positional argument so that the output can be restricted to matching clusters. Especially useful for `list --details` when there are lots of clusters.

bobvawter:
This change allows the list command to accept an optional regular expression to
match against cluster names.  A new --mine flag is added, which is a shorthand
for filtering by clusters owned by the current user.

Resolve #93

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/97)

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,6 +6,7 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"text/tabwriter"
@@ -53,6 +54,7 @@ var (
 	destroyAfter   time.Duration
 	extendLifetime time.Duration
 	listDetails    bool
+	listMine       bool
 	clusterType    = "cockroach"
 	secure         = false
 	nodeEnv        = "COCKROACH_ENABLE_RPC_COMPRESSION=false"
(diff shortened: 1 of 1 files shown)


petermattis:
:lgtm:

---

---

*[main.go, line 439 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/97#-L6Nk0DN1yx3ObKTsost:-L6Nk0DN1yx3ObKTsosu:b7m6ox4) ([raw file](https://github.com/cockroachdb/roachprod/blob/c9cd847611ca8e7a45684f1e0373cb3f3d6c0cd4/main.go#L439)):*
> ```Go
> 		var names []string
> 		for name, _ := range cloud.Clusters {
> 			if listPattern == nil || listPattern.MatchString(name) {
> ```

Nit: you could also always initialize `listPattern` and set it to `regexp.Compile(".*")` when you want to match everything.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/97#-:-L6NkCa4PSrMNtZDVvpy:bnfp4nl)*

bobvawter:

---

*[main.go, line 439 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/97#-L6Nk0DN1yx3ObKTsost:-L6No-BPsG169MEqzven:bt2vr1b) ([raw file](https://github.com/cockroachdb/roachprod/blob/c9cd847611ca8e7a45684f1e0373cb3f3d6c0cd4/main.go#L439)):*

Nit: you could also always initialize `listPattern` and set it to `regexp.Compile(".*")` when you want to match everything.
</blockquote></details>

Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/97)*

PR #108, opened 2018-03-01
Deduplicate gc notifications

### LINKED ISSUE #105: quiet the periodic slack messages
https://github.com/cockroachdb/roachprod/pull/94 replaced the `roachprod gc` email notifications with slack notifications. It also removed `--tracking-file` which was being used to not repeatedly send notifications about clusters. Experience with the slack notifications over the past day is that they are too noisy. In particular, the notifications are often repeating the same info over and over. It should be straightforward to only send a notification if it differs from the previous notification. That is, we re-add something akin to `--tracking-file` bug store the previous notification that was sent to the status channel as well as the per-user notifications. I'm imaging json-encoding something like:

```
type Status struct {
  good    []*CloudCluster
  warn    []*CloudCluster
  destroy []*CloudCluster
}

type Notifications struct {
  All Status
  Users map[string]*Status
}
```

I'm not too particular about the actual mechanism, so feel free to ignore the suggestion if there is a better/easier way.

bobvawter:
This change computes a hash value for each (channel, status) pair to debounce
notifications.

Resolves #105

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/108)

### DIFF
--- a/cloud/gc.go
+++ b/cloud/gc.go
@@ -1,8 +1,13 @@
 package cloud
 
 import (
+	"encoding/base64"
 	"fmt"
+	"hash/fnv"
+	"io/ioutil"
 	"log"
+	"os"
+	"path"
 	"sort"
 	"strings"
 	"time"
@@ -27,6 +32,33 @@ func (s *status) add(c *CloudCluster, now, destroyDeadline time.Time) {
 	} else {
 		s.warn = append(s.warn, c)
 	}
+}
+
+// messageHash computes a base64-encoded hash value to show whether
+// or not two status values would result in a duplicate
+// notification to a user.
+func (s *status) notificationHash() string {
+	// Use stdlib hash function, since we don't need any crypto guarantees
+	hash := fnv.New32a()
+
+	for i, list := range [][]*CloudCluster{s.good, s.warn, s.destroy} {
+		hash.Write([]byte{byte(i)})
+
+		var data []string
+		for _, c := range list {
+			// Deduplicate by cluster name and expiration time
+			data = append(data, fmt.Sprintf("%s %s", c.Name, c.ExpiresAt()))
+		}
+		// Ensure results are stable
+		sort.Strings(data)
+
+		for _, d := range data {
+			hash.Write([]byte(d))
+		}
+	}
+
+	bytes := hash.Sum(make([]byte, 0, hash.BlockSize()))
+	return base64.StdEncoding.EncodeToString(bytes)
 }
 
 func makeSlackClient() *slack.Client {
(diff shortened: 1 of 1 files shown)


cloud/gc.go
```diff
@@ -80,6 +112,18 @@ func postStatus(
 		return
 	}
 
+	// Debounce messages, unless we have badVMs since that indicates
+	// a problem that needs manual intervention
+	if len(badVMs) == 0 {
+		send, err := shouldSend(channel, s)
+		if err != nil {
+			log.Printf("unable to deduplicate notification: %s", err.Error())
```
benesch: tiny optional nit: `%s` will automatically call `.Error()` on errors

petermattis:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[cloud/gc.go, line 228 at r2](https://reviewable.io:443/reviews/cockroachdb/roachprod/108#-L6XWAiUIYaFwSuuCOOA:-L6XWAiVHEq5hWFLyW2d:b6yxzc3) ([raw file](https://github.com/cockroachdb/roachprod/blob/708d4bc14ed74d229ba0c447ce22b5fcb5398f0a/cloud/gc.go#L228)):*
> ```Go
> 		return true, err
> 	}
> 	hashPath := os.ExpandEnv(path.Join(hashDir, "notification-"+channel))
> ```

Is the `ExpandEnv` necessary here? Seems redundant. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/108#-:-L6XWMPU29l1tPIfifWI:bnfp4nl)*

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[cloud/gc.go, line 224 at r2](https://reviewable.io:443/reviews/cockroachdb/roachprod/108#-L6XWo-2mtjTkGc4hK8b:-L6XWo-2mtjTkGc4hK8c:b-b2yc2m) ([raw file](https://github.com/cockroachdb/roachprod/blob/708d4bc14ed74d229ba0c447ce22b5fcb5398f0a/cloud/gc.go#L224)):*
> ```Go
> // advisory; the boolean value is always a reasonable behavior.
> func shouldSend(channel string, status *status) (bool, error) {
> 	hashDir := os.ExpandEnv(path.Join("${HOME}", ".roachprod", "slack"))
> ```

FYI, `path` != `filepath`. The latter is what you want to use. Very confusing. I've changed this in my expiration PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/108)*

PR #111, opened 2018-03-01
Fix `status` on linux boxes

petermattis:
The `status` command was not displaying the process pid because `ps -o
ucomm` differs between OS X and Linux. Restructured that `status` shell
pipeline to avoid the discrepancy.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/111)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -166,7 +166,7 @@ func (c *SyncedCluster) Status() {
 
 		binary := cockroachNodeBinary(c, c.Nodes[i])
 		cmd := fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


benesch:
Forgive my ignorance, but how does reordering the pipeline fix things?

petermattis:
```
~ ps axeww -o ucomm -o pid -o command | grep cockroach
cockroach XDG_S 20444 ./cockroach start ...
```

```
~ ps axeww -o pid -o ucomm -o command | grep cockroach
20444 cockroach XDG_S ./cockroach
```

I want the binary name (sans directory) and the pid. Strangely, `ps axeww -o ucomm -o pid` doesn't print that XDG_S garbage, but it also doesn't show me the environment which is something else we need.

PR #112, opened 2018-03-02
ssh: permit multi-node ssh with iterm2 splits

jordanlewis:
This commit uses some AppleScript and iTerm2 magic to automatically pop
open a window with one split per node, and re-invoke roachprod ssh in
all of them. This allows you to have an ssh connection open to all of
the nodes in a roachprod cluster at once.

Then, you can use iTerm2's type-to-all-panes feature to interact with
all of the nodes at once (Shell->Broadcast Input->Broadcast Input to All
Panes in Current Tab, or Cmd-Ctrl-I).

Here's a gif of what it looks like:

https://gfycat.com/gifs/detail/LightheartedFavoriteFurseal

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/112)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -708,8 +708,13 @@ func (c *SyncedCluster) pgurls(nodes []int) map[int]string {
 }
 
 func (c *SyncedCluster) Ssh(sshArgs, args []string) error {
-	if len(c.Nodes) != 1 {
-		return fmt.Errorf("invalid number of nodes for ssh: %d", len(c.Nodes))
+	if len(c.Nodes) != 1 && len(args) == 0 {
+		// If trying to ssh to more than 1 node and the ssh session is interative,
+		// try sshing with an iTerm2 split screen configuration.
+		sshed, err := maybeSplitScreenSSHITerm2(c)
+		if sshed {
+			return err
+		}
 	}
 
 	allArgs := []string{
--- a/install/iterm2.go
+++ b/install/iterm2.go
@@ -1,0 +1,69 @@
+package install
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+)
+
+const splitScript = `
+ if (path to frontmost application as text) does not contain "iTerm" then
+        return "Can't multi-ssh without iTerm 2 as the active terminal."
+ end if
+ tell application "iTerm2"
+        activate
+        tell (create window with default profile)
+            set pane_count to %d
+            set col_count to round ((pane_count) ^ 0.5) rounding up
+            set row_count to round ((pane_count) / col_count) rounding up
+            repeat row_count - 1 times
+                tell current session
+                    split horizontally with default profile
+                end tell
+            end repeat
+            set cur to row_count
+            repeat row_count times
+                repeat col_count - 1 times
+                    if cur is equal to (pane_count) then
+                        exit repeat
+                    end if
+                    tell current session
+                        split vertically with default profile
+                    end tell
+                    set cur to cur + 1
+                end repeat
+                tell application "System Events" to tell process "iTerm2" to click menu item "Select Pane Below" of menu 1 of menu item "Select Split Pane" of menu 1 of menu bar item "Window" of menu bar 1
+            end repeat
+            tell current tab
+                set node_list to {%s}
+                repeat with i from 1 to pane_count
+                    tell item i of sessions to write text "roachprod ssh %s:" & item i of node_list
+                end repeat
+            end tell
+        end tell
+    end tell
+return`
+
+// maybeSplitScreenSSHITerm2 sshs to all of the specified nodes in the
+// SyncedCluster in an iTerm2 split-screen configuration if possible. It
+// returns true in the first position if it went through with the split-screen
+// ssh.
+func maybeSplitScreenSSHITerm2(c *SyncedCluster) (bool, error) {
+	osascriptPath, err := exec.LookPath("osascript")
+	if err != nil {
+		return false, err
+	}
+	nodeStrings := make([]string, len(c.Nodes))
+	for i, nodeID := range c.Nodes {
+		nodeStrings[i] = fmt.Sprint(nodeID)
+	}
+	script := fmt.Sprintf(splitScript, len(c.Nodes), strings.Join(nodeStrings, ","), c.Name)
+	allArgs := []string{
+		"osascript",
+		"-e",
+		script,
+	}
+	return true, syscall.Exec(osascriptPath, allArgs, os.Environ())
+}


jordanlewis:
Previously this invocation would just error out, so it wouldn't be breaking anyone's workflow. That being said it now does nothing if iTerm isn't the invoking terminal.

Out of curiosity what terminal do you use? @justinj wanted this to use `tmux` to split, so if people care I could add that at some point. Personally I like using the iTerm native splits better.

a-robinson:
That gif is both terrifying and incredible at the same time

petermattis:
:lgtm:

Like @benesch I feel this should be enabled by a flag. The previous behavior of erroring out on multiple nodes has actually been useful to me when developing `roachtest`-based tests. I wouldn't have wanted multiple ssh terminals opened. That said, I might change my mind when I actually use this. 

---

---

*[install/cluster_synced.go, line 733 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/112#-L6b5qrJytrqzF2AKqrC:-L6b5qrJytrqzF2AKqrD:bwdgkrc) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L733)):*
> ```Go
> 			script,
> 		}
> 		osascriptPath, err := exec.LookPath("osascript")
> ```

Nit: this should be the first thing you check in this if-block. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/112#-:-L6b5xonDY9xnUV5Hw8R:bagt3rp)*

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[install/cluster_synced.go, line 722 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/112#-L6b6ISDMXj4P_pG4Uyr:-L6b6ISDMXj4P_pG4Uys:b-5crhmv) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L722)):*
> ```Go
>         end tell
>     end tell
> return`
> ```

Can we move this to `iterm2.go`? Perhaps the code in `SyncedCluster.Ssh` as well. I'm thinking of something like `maybeSshIterm2()` or something like that.

---

*[install/cluster_synced.go, line 727 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/112#-L6b6X1u786Kj4cg98SB:-L6b6X1u786Kj4cg98SC:b-dn459m) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L727)):*
> ```Go
> 	if len(c.Nodes) != 1 {
> 		cmd := fmt.Sprintf(`"roachprod ssh %s:" & x`, c.Name)
> 		script := fmt.Sprintf(splitScript, len(c.Nodes), cmd)
> ```

I don't think this is selecting the right nodes. Consider `roachprod ssh local:1,4`. `len(c.Nodes)` will be 2 in that case, and I think you're currently translating that into `roachprod ssh local:1 and `roachprod ssh local:2`, while that second command should really be `roachprod ssh local:4`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/112)*

bdarnell:
FWIW, I don't use iterm2 either (just the regular apple terminal and emacs shells). I also use tmux/screen, although I'm much less interested into a "type into a bunch of windows simultaneously" solution than a "run this command everywhere and collect the outputs" like we have with `crl-ssh`. 

---

Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[install/cluster_synced.go, line 725 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/112#-L6bH3srszWW9Blx_hlJ:-L6bH3ss2GQBc1F1OU3c:bfazsnx) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L725)):*
> ```Go
> 
> func (c *SyncedCluster) Ssh(sshArgs, args []string) error {
> 	if len(c.Nodes) != 1 {
> ```

If there are args, we should run that command on all nodes instead of doing all this iterm magic. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/112#-:-L6bHJIwaa2kUFZ5PeHC:bj2nsoj)*

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[install/cluster_synced.go, line 725 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112#-L6bH3srszWW9Blx_hlJ:-L6bIfdMCe6MnSFaZvB8:bvuk1c2) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L725)):*

If there are args, we should run that command on all nodes instead of doing all this iterm magic. 
</blockquote></details>

What do you do with the output from each command? See also `roachprod run` which has significant overlap here. I didn't take sufficient time to merge the two commands, but they really deserve to be merged. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112)*

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[install/cluster_synced.go, line 725 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112#-L6bH3srszWW9Blx_hlJ:-L6bJ8ob0xA_njrV103a:b-dvyxvc) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L725)):*

What do you do with the output from each command? See also `roachprod run` which has significant overlap here. I didn't take sufficient time to merge the two commands, but they really deserve to be merged. 
</blockquote></details>

Ah, didn't realize there was a separate `run` command, which is more or less what i want. My preference (assuming parallelism) is to prefix each line with the hostname (or node number) and send it to the current stdout by default (interleaved line by line is better than buffering everything to display non-interleaved output). An option to send the output to a file named for each node would be nice but less important IMO.  

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112)*

petermattis:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[install/cluster_synced.go, line 725 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112#-L6bH3srszWW9Blx_hlJ:-L6bKT2u-ca1or_Wkax4:b3ve4b3) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L725)):*

Ah, didn't realize there was a separate `run` command, which is more or less what i want. My preference (assuming parallelism) is to prefix each line with the hostname (or node number) and send it to the current stdout by default (interleaved line by line is better than buffering everything to display non-interleaved output). An option to send the output to a file named for each node would be nice but less important IMO.  
</blockquote></details>

`run` currently buffers the output and display it in node order. For my use cases that has been appropriate. I can imagine other use cases which desire different behavior. Adding flags seems appropriate. PRs are welcome.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112)*

jordanlewis:
Going to leave this available without a flag. If this actually becomes a usability issue for people, let's hide it behind a flag at that point.

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[install/cluster_synced.go, line 722 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112#-L6b6ISDMXj4P_pG4Uyr:-L9RyKNe3VOQlBCXTbz_:b-896fix) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L722)):*

Can we move this to `iterm2.go`? Perhaps the code in `SyncedCluster.Ssh` as well. I'm thinking of something like `maybeSshIterm2()` or something like that.
</blockquote></details>

Done.

---

*[install/cluster_synced.go, line 725 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112#-L6bH3srszWW9Blx_hlJ:-L9RyOC0EOg7Z5q51hoO:b-896fix) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L725)):*

`run` currently buffers the output and display it in node order. For my use cases that has been appropriate. I can imagine other use cases which desire different behavior. Adding flags seems appropriate. PRs are welcome.
</blockquote></details>

Done.

---

*[install/cluster_synced.go, line 727 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112#-L6b6X1u786Kj4cg98SB:-L6b9pWsek4aoAp8jp_g:b-896fix) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L727)):*

I don't think this is selecting the right nodes. Consider `roachprod ssh local:1,4`. `len(c.Nodes)` will be 2 in that case, and I think you're currently translating that into `roachprod ssh local:1 and `roachprod ssh local:2`, while that second command should really be `roachprod ssh local:4`.
</blockquote></details>

Done.

---

*[install/cluster_synced.go, line 733 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112#-L6b5qrJytrqzF2AKqrC:-L9RyMdl94U7POce8E5e:b-896fix) ([raw file](https://github.com/cockroachdb/roachprod/blob/fe2c76d612b7ae680fd8899694b7608642047eb4/install/cluster_synced.go#L733)):*

Nit: this should be the first thing you check in this if-block. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/112#-:-L9RyOtA3uVk2FNrrC59:b-1yb28k)*

PR #113, opened 2018-03-02
Improve error message when unable to ssh to localhost

### LINKED ISSUE #98: `roachprod local` ssh failure requires a better error message
There was a cryptic error message saying 'failed' and nothing more. I needed to install openssh-server, and create a new passphrase-less rsa keypair and add it to my `authorized_keys` to get it to work.

bobvawter:
Resolves #98

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/113)

### DIFF
--- a/ssh/ssh.go
+++ b/ssh/ssh.go
@@ -153,6 +153,10 @@ func NewSSHSession(user, host string) (*ssh.Session, error) {
 		var err error
 		client.Client, _, err = newSSHClient(user, host)
 		if err != nil {
+			if host == "127.0.0.1" || host == "localhost" {
+				err = errors.Wrap(err, "could not ssh to localhost: "+
+					"ensure that you have password-less sshd set up")
+			}
 			return nil, err
 		}
 	}


petermattis:
:lgtm:

I wonder if we should change `local` to avoid using `ssh`. I made the decision to use `ssh` because it seemed expedient, but at least 3 different folks have encountered this issue when running `roachprod` for the first time. Thoughts?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/113#-:-L6by89FKUXd_3Gw0zp9:bm32qjl)*

bobvawter:
I can see hoisting out a `ClusterSession` type vended by `SyncedCluster` that provides most of the existing `ssh.Session` behaviors, with special-case handling for the local machine.  That said, the current approach of ssh'ing to the local machine requires less development work since there's only ever one code path to accomplish a task.  I think asking a `roachprod` user to start an ssh server is entirely straightforward and well within expected competencies.

petermattis:
> I think asking a roachprod user to start an ssh server is entirely straightforward and well within expected competencies.

Roger. Thanks for the thoughts. Let's not worry about this now.

PR #114, opened 2018-03-02
Force aws output to json

bobvawter:
It's possible for a user to configure their local environment such that the aws
tool will output in textual or tabular formats.  This change always sets the
--output option to json.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/114)

### DIFF
--- a/vm/aws/support.go
+++ b/vm/aws/support.go
@@ -54,6 +54,8 @@ func runCommand(args []string) error {
 
 // runJSONCommand invokes an aws command and parses the json output.
 func runJSONCommand(args []string, parsed interface{}) error {
+	// force json output in case the user has overridden the default behavior
+	args = append(args[:len(args):len(args)], "--output", "json")
 	cmd := exec.Command("aws", args...)
 
 	rawJSON, err := cmd.Output()


petermattis:
:lgtm:

---

---

*[vm/aws/support.go, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/114#-L6cKj15bq3Aevr8O6kb:-L6cKj15bq3Aevr8O6kc:b9ocm5h) ([raw file](https://github.com/cockroachdb/roachprod/blob/112811f872848eeb008f4413225ae881970ec9ca/vm/aws/support.go#L60)):*
> ```Go
> 	argCopy := make([]string, len(args), len(args)+2)
> 	copy(argCopy, args)
> 	argCopy = append(argCopy, "--output", "json")
> ```

You could perform the above in a single step:

```
argCopy := append(args[:len(args):len(args)], "--output", "json")
```

If you haven't seen the triple-index syntax before, the second index is the length of the slice and the third is the capacity. By setting the capacity to the length we're ensuring that the `append` will have to allocate a new slice. Perhaps a bit more subtle that your code, but it is good to be familiar with slide length vs capacity.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/114#-:-L6cL4XSQh_3ZLbCrSbQ:bnfp4nl)*

tschottdorf:
Thanks @bobvawter! Can confirm that this fixes it for me even though I have `text` as my default output format for whatever reason.

PR #122, opened 2018-03-05
Don't leak VMs if create fails

### LINKED ISSUE #121: Nodes are leaked when roachprod create fails
During the nightly tests, if we hit our ssd quota, roachprod create will return a failed exit code, which correctly fails the test, but it doesn't first tear down any instances it does create. They get cleaned up eventually, but in the meantime are costing us money and using precious quota.

benesch:
Fix #121.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/122)

### DIFF
--- a/main.go
+++ b/main.go
@@ -261,11 +261,17 @@ Local Clusters
 		}
 
 		fmt.Printf("Creating cluster %s with %d nodes\n", clusterName, numNodes)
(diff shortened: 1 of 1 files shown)


petermattis:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/122#-:-L6rvWB8FsefBMTadAcc:bnfp4nl)*

Issue #79, opened 2018-02-22
Remove test, web and dump commands


petermattis:
The `test`, `web` and `dump` commands were inherited from `roachperf`. In the current `roachprod/roachtest` world, these commands make more sense in `roachtest`. The tests deserve to be `roachtest`-style tests. The `web` and `dump` commands could be migrated as-is or perhaps moved to a different tool.

petermattis:
Removing these commands can lead to some other cleanup, such as getting rid of `SyncedCluster.{LoadGen,RunLoad,CockroachVersions}`.

PR #126, opened 2018-03-06
Provide more detail on errors

### LINKED ISSUE #125: failed start command should be more helpful
`roachprod start arjun-tpcc-checks:1-3` returns
```
arjun-tpcc-checks: starting 3/3
0: Process exited with status 1: 
1: Process exited with status 1: 
2: Process exited with status 1: 
2018/03/06 16:52:11 command failed
```

This confused me, because this was not clear that it is the `cockroach start` command that `roachprod` was running that failed. We should be more clear as to what failed, as I spent some time trying to figure out why `roachprod` exited.

petermattis:
When an error occurs exec'ing various processes, include both the output
and the error in the message. For example:

```
~ roachprod start local
local: starting 2/2
0: bash: /Users/pmattis/local/1/cockroach: is a directory
: Process exited with status 126:
1: bash: /Users/pmattis/local/2/cockroach: is a directory
: Process exited with status 126:
2018/03/06 18:43:27 command failed
```

Fixes #125

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/126)

### DIFF
--- a/install/cockroach.go
+++ b/install/cockroach.go
@@ -14,6 +14,7 @@ import (
 	"github.com/cockroachdb/roachprod/config"
 	"github.com/cockroachdb/roachprod/ssh"
 	"github.com/hashicorp/go-version"
+	"github.com/pkg/errors"
 )
 
 var StartOpts struct {
@@ -69,7 +70,7 @@ func getCockroachVersion(c *SyncedCluster, i int, host, user string) (*version.V
 	cmd := cockroachNodeBinary(c, i) + " version"
 	out, err := session.CombinedOutput(cmd)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "~ %s\n%s", cmd, out)
 	}
 
 	matches := regexp.MustCompile(`(?m)^Build Tag:\s+(.*)$`).FindSubmatch(out)
@@ -131,7 +132,7 @@ func (r Cockroach) Start(c *SyncedCluster, extraArgs []string) {
 				c.Parallel("", len(nodes), 0, func(i int) ([]byte, error) {
 					var err error
 					ips[i], err = c.GetInternalIP(nodes[i])
-					return nil, err
+					return nil, errors.Wrapf(err, "IPs")
 				})
 			}
 
@@ -169,8 +170,8 @@ mkdir -p certs
 %[1]s cert create-node localhost %[2]s --certs-dir=certs --ca-key=certs/ca.key
 tar cvf certs.tar certs
 `, cockroachNodeBinary(c, 1), strings.Join(nodeNames, " "))
-				if _, err := session.CombinedOutput(cmd); err != nil {
-					msg = err.Error()
+				if out, err := session.CombinedOutput(cmd); err != nil {
+					msg = fmt.Sprintf("%s: %v", out, err)
 				}
 				return nil, nil
 			})
@@ -225,8 +226,10 @@ tar cvf certs.tar certs
 					cmd = fmt.Sprintf(`cd ${HOME}/local/%d ; `, nodes[i])
 				}
 				cmd += `tar xf -`
-				_, err = session.CombinedOutput(cmd)
-				return nil, err
+				if out, err := session.CombinedOutput(cmd); err != nil {
+					return nil, errors.Wrapf(err, "~ %s\n%s", cmd, out)
+				}
+				return nil, nil
 			})
 		}
 	}
@@ -314,11 +317,17 @@ tar cvf certs.tar certs
 		// NB: this is awkward as when the process fails, the test runner will show an
 		// unhelpful empty error (since everything has been redirected away). This is
 		// unfortunately equally awkward to address.
-		cmd := "mkdir -p " + logDir + "; " + c.Env +
+		cmd := "mkdir -p " + logDir + "; " +
 			fmt.Sprintf(" export ROACHPROD=%d && ", nodes[i]) +
-			binary + " start " + strings.Join(args, " ") +
-			" >> " + logDir + "/cockroach.stdout 2>> " + logDir + "/cockroach.stderr"
-		return session.CombinedOutput(cmd)
+			c.Env + " " + binary + " start " + strings.Join(args, " ") +
+			" >> " + logDir + "/cockroach.stdout 2>> " + logDir + "/cockroach.stderr" +
+			" || (x=$?; cat " + logDir + "/cockroach.stderr; exit $x)"
+		if out, err := session.CombinedOutput(cmd); err != nil {
+			return nil, errors.Wrapf(err, "~ %s\n%s", cmd, out)
+		}
+		// NB: if cockroach started successfully, we ignore the output as it is
+		// some harmless start messaging.
+		return nil, nil
 	})
 
 	if bootstrapped {
@@ -349,10 +358,9 @@ SET CLUSTER SETTING cluster.organization = 'Cockroach Labs - Production Testing'
 SET CLUSTER SETTING enterprise.license = '%s';"`, license)
 			out, err := session.CombinedOutput(cmd)
 			if err != nil {
-				msg = "while setting cluster settings: " + err.Error()
-			} else {
-				msg = strings.TrimSpace(string(out))
+				return nil, errors.Wrapf(err, "~ %s\n%s", cmd, out)
 			}
+			msg = strings.TrimSpace(string(out))
 			return nil, nil
 		})
 
@@ -407,6 +415,7 @@ func (r Cockroach) SQL(c *SyncedCluster, args []string) error {
 	type result struct {
 		node   int
 		output string
+		err    error
 	}
 	resultChan := make(chan result, len(c.Nodes))
 
@@ -428,8 +437,7 @@ func (r Cockroach) SQL(c *SyncedCluster, args []string) error {
 
 		out, err := session.CombinedOutput(cmd)
 		if err != nil {
-			resultChan <- result{node: c.Nodes[i], output: fmt.Sprintf("err=%s,out=%s", err, out)}
-			return out, err
+			return nil, errors.Wrapf(err, "~ %s\n%s", cmd, out)
 		}
 
 		resultChan <- result{node: c.Nodes[i], output: string(out)}
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -182,12 +182,11 @@ fi
 		out, err := session.CombinedOutput(cmd)
 		var msg string
 		if err != nil {
-			msg = err.Error()
-		} else {
-			msg = strings.TrimSpace(string(out))
-			if msg == "" {
-				msg = "not running"
-			}
+			return nil, errors.Wrapf(err, "~ %s\n%s", cmd, out)
+		}
+		msg = strings.TrimSpace(string(out))
+		if msg == "" {
+			msg = "not running"
 		}
 		results[i] = msg
 		return nil, nil
@@ -380,7 +379,7 @@ func (c *SyncedCluster) CockroachVersions() map[string]int {
 		out, err := session.CombinedOutput(cmd)
 		var s string
 		if err != nil {
-			s = err.Error()
+			s = fmt.Sprintf("%s: %v", out, err)
 		} else {
 			s = strings.TrimSpace(string(out))
 		}
@@ -681,7 +680,7 @@ func (c *SyncedCluster) pgurls(nodes []int) map[int]string {
 	c.Parallel("", len(nodes), 0, func(i int) ([]byte, error) {
 		var err error
 		ips[i], err = c.GetInternalIP(nodes[i])
-		return nil, err
+		return nil, errors.Wrapf(err, "pgurls")
 	})
 
 	m := make(map[int]string, len(ips))
@@ -843,7 +842,7 @@ func (c *SyncedCluster) Parallel(display string, count, concurrency int, fn func
 	if len(failed) > 0 {
 		sort.Slice(failed, func(i, j int) bool { return failed[i].index < failed[j].index })
 		for _, f := range failed {
-			fmt.Fprintf(os.Stderr, "%d: %s: %s\n", f.index, f.err, f.out)
+			fmt.Fprintf(os.Stderr, "%d: %+v: %s\n", f.index, f.err, f.out)
 		}
 		log.Fatal("command failed")
 	}


petermattis:
Cc @arjunravinarayan

install/cockroach.go
```diff
@@ -69,7 +70,7 @@ func getCockroachVersion(c *SyncedCluster, i int, host, user string) (*version.V
 	cmd := cockroachNodeBinary(c, i) + " version"
 	out, err := session.CombinedOutput(cmd)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "%s", out)
```
nvanbenschoten: Can't these be replaced with `errors.Wrap(err, out)`? If so, slight preference everywhere to create strings with `errors.Wrap(err, out).Error()`.

petermattis: `out` is a `[]byte`. The default formatting for `[]byte` is unreadable. I could make this `errors.Wrap(err, string(out))`, though I prefer the explicitness of `errors.Wrapf`.

nvanbenschoten: In that case, I like this approach better too.

benesch:
```
0: bash: /Users/pmattis/local/1/cockroach: is a directory
: Process exited with status 126:
1: bash: /Users/pmattis/local/2/cockroach: is a directory
: Process exited with status 126:
2018/03/06 18:43:27 command failed
```

That's not very pretty output. I think printing `err: output` instead of `output: err` would go miles:

```
0: Process exited with status 126: bash: /Users/pmattis/local/1/cockroach: is a directory
1: Process exited with status 126: bash: /Users/pmattis/local/2/cockroach: is a directory
2018/03/06 18:43:27 command failed
```

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[install/cockroach.go, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/126#-L6xjIlLfN5YsISp91HM-r1-73:-L6ymJi7V4cYDZCuX6-n:b-oyjces) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L73)):*
<details><summary><i>Previously, nvanbenschoten (Nathan VanBenschoten) wrote…</i></summary><blockquote>

In that case, I like this approach better too.
</blockquote></details>

Per my root comment I think this would be better as `fmt.Errorf("%s: %s", err, strings.TrimSpace(out))`. 

---

*[install/cockroach.go, line 324 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/126#-L6ykTHgQ-netgis1vVG:-L6ykTHgQ-netgis1vVH:b-upz8yg) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L324)):*
> ```Go
> 			binary + " start " + strings.Join(args, " ") +
> 			" >> " + logDir + "/cockroach.stdout 2>> " + logDir + "/cockroach.stderr"
> 		return session.CombinedOutput(cmd)
> ```

Have you run this against a remote cluster where `cockroach start` fails? That's the case Arjun filed #125 about and I think it's the one case you don't handle here! It's also the hardest case; the output of `cockroach start` is swallowed by these redirections.

I think the simplest approach is to tack on make `cmd` look like this:

```
cockroach start --insecure .... > logs/cockroach.stdout 2> logs/cockroach.stderr || cat logs/cockroach.stdout logs/cockroach.stderr 2> /dev/null
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/126#-:-L6ylRZiVnAAK-kKAgIK:bgwd3uo)*

benesch:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[install/cockroach.go, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/126#-L6xjIlLfN5YsISp91HM-r1-73:-L6ymYiBw9RsmotLMZAU:beqkw2r) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L73)):*
<details><summary><i>Previously, benesch (Nikhil Benesch) wrote…</i></summary><blockquote>

Per my root comment I think this would be better as `fmt.Errorf("%s: %s", err, strings.TrimSpace(out))`. 
</blockquote></details>

Make that `fmt.Errorf("%s: %s", err, bytes.TrimSpace(out))`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/126)*

tschottdorf:
Thanks for doing this!

---

Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[install/cockroach.go, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/126#-L6xjIlLfN5YsISp91HM-r1-73:-L6zKoxMmUmhXmBpDx4_:bn7r5tv) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L73)):*
<details><summary><i>Previously, benesch (Nikhil Benesch) wrote…</i></summary><blockquote>

Make that `fmt.Errorf("%s: %s", err, bytes.TrimSpace(out))`.
</blockquote></details>

I think it's worth sticking with `pkg/errors` because those contain [stack traces](https://godoc.org/github.com/pkg/errors#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper) which I also think we should leverage; it's often annoying to have to grep your way to the source of a particular error, and sometimes it's even impossible.

Any error in `roachtest` is usually fatal, so we might as well `%+v` them, and then the stack trace will go a long way. (Not suggesting that kind of work be done in this PR).

---

*[install/cockroach.go, line 230 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/126#-L6zLOkMw56X5TSnluoW:-L6zLOkMw56X5TSnluoX:bbek3bk) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L230)):*
> ```Go
> 				cmd += `tar xf -`
> 				if out, err := session.CombinedOutput(cmd); err != nil {
> 					return nil, errors.Wrapf(err, "%s", out)
> ```

For greppability, it would help to do `errors.Wrapf(err, "starting node: %s", out)`.

---

*[install/cockroach.go, line 324 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/126#-L6ykTHgQ-netgis1vVG:-L6zLl_Tn24dH-OLCeZ2:b-4lf094) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L324)):*
<details><summary><i>Previously, benesch (Nikhil Benesch) wrote…</i></summary><blockquote>

Have you run this against a remote cluster where `cockroach start` fails? That's the case Arjun filed #125 about and I think it's the one case you don't handle here! It's also the hardest case; the output of `cockroach start` is swallowed by these redirections.

I think the simplest approach is to tack on make `cmd` look like this:

```
cockroach start --insecure .... > logs/cockroach.stdout 2> logs/cockroach.stderr || cat logs/cockroach.stdout logs/cockroach.stderr 2> /dev/null
```
</blockquote></details>

@benesch but that'll change the exit code to success, won't it? (I wrote the `NB:` above and would love to see this fixed as it's about as opaque as it gets). Do we really have to resort to something like below?

```
./cockroach start ... || (x=$?; cat ... && exit $?)
```

---

*[install/cockroach.go, line 355 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/126#-L6zMDuRH49jiHYa7dM3:-L6zMDuS2cRAvx8XxymY:b-27ies2) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L355)):*
> ```Go
> 			out, err := session.CombinedOutput(cmd)
> 			if err != nil {
> 				msg = fmt.Sprintf("%s: %v", out, err)
> ```

The old variant was a lot more greppable, I'd try to maintain that.

---

*[install/cockroach.go, line 434 at r1](https://reviewable.io:443/reviews/cockroachdb/roachprod/126#-L6zMgqKAiLdIH4c20Lx:-L6zMgqKAiLdIH4c20Ly:bdk1mt5) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L434)):*
> ```Go
> 		out, err := session.CombinedOutput(cmd)
> 		if err != nil {
> 			resultChan <- result{node: c.Nodes[i], output: fmt.Sprintf("%s: %v", out, err)}
> ```

Here too it would help to make it greppable. Maybe inserting an `errors.Wrapf(err, "SQL on node %", i)` above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/roachprod/126#-:-L6zMxhjxkUgge58Bu98:bxgavm9)*

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[install/cockroach.go, line 324 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6ykTHgQ-netgis1vVG:-L7-eLydCC4ElqZGOPaO:b-v6vkae) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L324)):*
<details><summary><i>Previously, tschottdorf (Tobias Schottdorf) wrote…</i></summary><blockquote>

@benesch but that'll change the exit code to success, won't it? (I wrote the `NB:` above and would love to see this fixed as it's about as opaque as it gets). Do we really have to resort to something like below?

```
./cockroach start ... || (x=$?; cat ... && exit $?)
```
</blockquote></details>

The earlier `getCockroachVersion` call will detect if the binary isn't runnable at all. But you have a good point that I didn't completely solve the problem. For example, `roachprod start local -a "--foo"` will still print a terse message. I've fixed this by removing the stdout and stderr redirection. That was a historical artifact from `roachperf` when I was using `--logtostderr`. Now that I write this, I'm worried that not redirecting stdout/stderr can lead to problems. Hmm, maybe I should take @tschottdorf's approach. 

---

*[install/cockroach.go, line 355 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6zMDuRH49jiHYa7dM3:-L7-fIbJ5dUTpxgO9HON:b-gs690) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L355)):*
<details><summary><i>Previously, tschottdorf (Tobias Schottdorf) wrote…</i></summary><blockquote>

The old variant was a lot more greppable, I'd try to maintain that.
</blockquote></details>

Good point. I've tweaked this to include the command that failed which provides both greppability and additional detail.

---

*[install/cockroach.go, line 434 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6zMgqKAiLdIH4c20Lx:-L7-gbS_Ac4Kt3Ut2qGn:byb7lit) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L434)):*
<details><summary><i>Previously, tschottdorf (Tobias Schottdorf) wrote…</i></summary><blockquote>

Here too it would help to make it greppable. Maybe inserting an `errors.Wrapf(err, "SQL on node %", i)` above.
</blockquote></details>

Similar to above, I've included the command that failed in the message. I've also fixed this so that it actually displays this output

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126)*

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[install/cockroach.go, line 324 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6ykTHgQ-netgis1vVG:-L7-n4djEmWvZKB1WRgF:b-mo1v3k) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L324)):*

The earlier `getCockroachVersion` call will detect if the binary isn't runnable at all. But you have a good point that I didn't completely solve the problem. For example, `roachprod start local -a "--foo"` will still print a terse message. I've fixed this by removing the stdout and stderr redirection. That was a historical artifact from `roachperf` when I was using `--logtostderr`. Now that I write this, I'm worried that not redirecting stdout/stderr can lead to problems. Hmm, maybe I should take @tschottdorf's approach. 
</blockquote></details>

Ok, I've added @tschottdorf's nugget and verified it works. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126)*

tschottdorf:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[install/cluster_synced.go, line 382 at r2](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L70AA1mEIUAa5OU_XTG:-L70AA1mEIUAa5OU_XTH:b-938ypt) ([raw file](https://github.com/cockroachdb/roachprod/blob/5f662ff9da41b72604aec91331bc1f62cc137b35/install/cluster_synced.go#L382)):*
> ```Go
> 		var s string
> 		if err != nil {
> 			s = fmt.Sprintf("%s: %v", out, err)
> ```

Did you want to `~ ` this as well?

---

*[install/cockroach.go, line 324 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6ykTHgQ-netgis1vVG:-L709LzD0I27RMwv9AzI:bz4xbps) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L324)):*
> The earlier getCockroachVersion call will detect if the binary isn't runnable at all.

Hmm, is that new? That didn't work for me a few days ago and Arjun's issue also seems to indicate that it didn't work for him yesterday.

I'm not seeing the nugget, did you forget to push?

---

*[install/cockroach.go, line 355 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6zMDuRH49jiHYa7dM3:-L709dVv4tUfps1yEaU0:bdrdjeo) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L355)):*

Good point. I've tweaked this to include the command that failed which provides both greppability and additional detail.
</blockquote></details>

This should do the trick, but it's not quite greppable yet, right? What would you grep for? The best I can do now is search for "~ " and then figure out which of the lines it is.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126)*

petermattis:
Review status: 1 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[install/cluster_synced.go, line 382 at r2](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L70AA1mEIUAa5OU_XTG:-L70BHvA0_BiDjZ51rte:b-wprzto) ([raw file](https://github.com/cockroachdb/roachprod/blob/5f662ff9da41b72604aec91331bc1f62cc137b35/install/cluster_synced.go#L382)):*
<details><summary><i>Previously, tschottdorf (Tobias Schottdorf) wrote…</i></summary><blockquote>

Did you want to `~ ` this as well?
</blockquote></details>

No, that was intentional to keep the output smaller. This code (`CockroachVersion`) is going to die very soon. See #79.

---

*[install/cockroach.go, line 73 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6xjIlLfN5YsISp91HM-r1-73:-L70BVB2BEk67jaCiRth:bh87o3k) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L73)):*
<details><summary><i>Previously, tschottdorf (Tobias Schottdorf) wrote…</i></summary><blockquote>

I think it's worth sticking with `pkg/errors` because those contain [stack traces](https://godoc.org/github.com/pkg/errors#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper) which I also think we should leverage; it's often annoying to have to grep your way to the source of a particular error, and sometimes it's even impossible.

Any error in `roachtest` is usually fatal, so we might as well `%+v` them, and then the stack trace will go a long way. (Not suggesting that kind of work be done in this PR).
</blockquote></details>

`SyncedCluster.Parallel` formats the error with `%+v` now. The stack traces are usually quite small.

---

*[install/cockroach.go, line 230 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6zLOkMw56X5TSnluoW:-L70BedzD7NSDyEjz3zm:bt3t5z0) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L230)):*
<details><summary><i>Previously, tschottdorf (Tobias Schottdorf) wrote…</i></summary><blockquote>

For greppability, it would help to do `errors.Wrapf(err, "starting node: %s", out)`.
</blockquote></details>

See other comment about grepping for the command text (i.e. `tar xf`).

---

*[install/cockroach.go, line 324 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6ykTHgQ-netgis1vVG:-L70Ah0S3vYnWvopdkGk:btm86fu) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L324)):*
Yes, I forgot to push. Done.

> Hmm, is that new? That didn't work for me a few days ago and Arjun's issue also seems to indicate that it didn't work for him yesterday.

No, it's not new, but the error reporting is. What Arjun was seeing was an execution failure of `cockroach version`. Prior to this PR nothing interesting was output. Now, `getCockroachVersion()` includes an expanded error message. 

---

*[install/cockroach.go, line 355 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6zMDuRH49jiHYa7dM3:-L70B8476xhRK5SE82qp:b-1n0roy) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L355)):*
<details><summary><i>Previously, tschottdorf (Tobias Schottdorf) wrote…</i></summary><blockquote>

This should do the trick, but it's not quite greppable yet, right? What would you grep for? The best I can do now is search for "~ " and then figure out which of the lines it is.
</blockquote></details>

The error message contains the complete command (e.g. `SET CLUSTER SETTING`) which you can then grep for. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126)*

tschottdorf:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[install/cockroach.go, line 324 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-L6ykTHgQ-netgis1vVG:-L70G906-6wnEzRSuzhG:b6ibvbl) ([raw file](https://github.com/cockroachdb/roachprod/blob/060b01047b2c25715b9b30748f171659993a51ff/install/cockroach.go#L324)):*

Yes, I forgot to push. Done.

> Hmm, is that new? That didn't work for me a few days ago and Arjun's issue also seems to indicate that it didn't work for him yesterday.

No, it's not new, but the error reporting is. What Arjun was seeing was an execution failure of `cockroach version`. Prior to this PR nothing interesting was output. Now, `getCockroachVersion()` includes an expanded error message. 
</blockquote></details>

Ah, TIL. I never even traced the error source to that.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126)*

tschottdorf:
:lgtm:

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/126#-:-L70GEPq2f7sptnju9oj:bnfp4nl)*

PR #139, opened 2018-03-29
run: use the expander in roachprod run

bdarnell:
The run and ssh commands were not as interchangeable as I had believed
when writing cockroachdb/cockroach#24322

Addresses part of #138

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/139)

PR #140, opened 2018-03-29
Combine the run and ssh commands

### LINKED ISSUE #138: Unify `roachprod run` and `roachprod ssh`
The `run` and `ssh` subcommands are almost but not quite identical: `ssh` can be interactive, but only runs against one node, while `run` is never interactive but supports parallel execution across multiple nodes. There are other subtle differences, like the fact that `ssh` runs the `expander` on its arguments while `run` does not. We should combine these into one command that supports multiple nodes in non-interactive mode (so we only have one interpretation of non-interactive commands) while also supporting interactive use on a single node.

petermattis:
Combine the `run` and `ssh` commands. The previous behavior of `ssh` is
accessed by targetting a single node or specifying no command to run. If
multiple nodes are specified, the previous behavior of `run` is
accessed. Both commands are still accessible (`ssh` is now an alias for
`run`).

Fixes #138

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/140)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -285,7 +285,16 @@ done
 }
 
 func (c *SyncedCluster) Run(w io.Writer, nodes []int, title, cmd string) error {
-	display := fmt.Sprintf("%s: %s", c.Name, title)
+	var e expander
+	cmd = e.expand(c, cmd)
+
+	// Stream output if we're running the command on only 1 node.
+	stream := len(nodes) == 1
+	var display string
+	if !stream {
+		display = fmt.Sprintf("%s: %s", c.Name, title)
+	}
+
 	errors := make([]error, len(nodes))
 	results := make([]string, len(nodes))
 	c.Parallel(display, len(nodes), 0, func(i int) ([]byte, error) {
@@ -302,6 +311,12 @@ func (c *SyncedCluster) Run(w io.Writer, nodes []int, title, cmd string) error {
 			nodeCmd = fmt.Sprintf("cd ${HOME}/local/%d ; %s", nodes[i], cmd)
 		}
 
+		if stream {
+			session.Stdout = w
+			errors[i] = session.Run(nodeCmd)
+			return nil, nil
+		}
+
 		out, err := session.CombinedOutput(nodeCmd)
 		msg := strings.TrimSpace(string(out))
 		if err != nil {
@@ -312,8 +327,10 @@ func (c *SyncedCluster) Run(w io.Writer, nodes []int, title, cmd string) error {
 		return nil, nil
 	})
 
-	for i, r := range results {
-		fmt.Fprintf(w, "  %2d: %s\n", nodes[i], r)
+	if !stream {
+		for i, r := range results {
+			fmt.Fprintf(w, "  %2d: %s\n", nodes[i], r)
+		}
 	}
 
 	for _, err := range errors {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -842,23 +842,27 @@ nodes.
...
 `,
-	Args: cobra.MinimumNArgs(2),
+	Args: cobra.MinimumNArgs(1),
 	Run: wrap(func(_ *cobra.Command, args []string) error {
 		c, err := newCluster(args[0], false /* reserveLoadGen */)
 		if err != nil {
 			return err
 		}
 
+		if len(args) == 1 || len(c.Nodes) == 1 {
```
bdarnell: I think this condition should be `if len(args) == 1`. If there's a non-interactive command, it should always go through `c.Run`, instead of switching based on whether it targets one or multiple nodes.

install/cluster_synced.go
```diff
@@ -705,12 +705,14 @@ func (c *SyncedCluster) Ssh(sshArgs, args []string) error {
 		"-o", "StrictHostKeyChecking=no",
 	}
 	allArgs = append(allArgs, sshArgs...)
+	if c.IsLocal() {
+		allArgs = append(allArgs,
+			"-t",
```
bdarnell: `-t` shouldn't be conditioned on whether it's local or not; it should be based on whether it's interactive. (or if you take my suggestion below, just remove `args` and then `roachprod ssh` is always interactive)

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[main.go, line 857 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/140#-L8mikwyQvaHKUGtbzf6-r1-857:-L8msgNv5UtzEF83PoT4:bplbdt3) ([raw file](https://github.com/cockroachdb/roachprod/blob/b8cd2ca2269720ec8dbbd6984cb4f9a12c6492ab/main.go#L857)):*

I think this condition should be `if len(args) == 1`. If there's a non-interactive command, it should always go through `c.Run`, instead of switching based on whether it targets one or multiple nodes. 
</blockquote></details>

`run` buffers the output while `ssh` streams it. That difference is important in some of my usage (e.g. `roachprod ssh peter-tpcc:1 -- ./tpcc ...`).

---

*[install/cluster_synced.go, line 710 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/140#-L8mikwyQvaHKUGtbzf5-r1-710:-L8mtZEQBwUqa6Ab4xBv:bl8lja6) ([raw file](https://github.com/cockroachdb/roachprod/blob/b8cd2ca2269720ec8dbbd6984cb4f9a12c6492ab/install/cluster_synced.go#L710)):*

`-t` shouldn't be conditioned on whether it's local or not; it should be based on whether it's interactive. (or if you take my suggestion below, just remove `args` and then `roachprod ssh` is always interactive)
</blockquote></details>

Good point. Fixed.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/140)*

bdarnell:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[main.go, line 857 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/140#-L8mikwyQvaHKUGtbzf6-r1-857:-L8muQwt7U6oq41gaGbo:bvkqgkb) ([raw file](https://github.com/cockroachdb/roachprod/blob/b8cd2ca2269720ec8dbbd6984cb4f9a12c6492ab/main.go#L857)):*

`run` buffers the output while `ssh` streams it. That difference is important in some of my usage (e.g. `roachprod ssh peter-tpcc:1 -- ./tpcc ...`).
</blockquote></details>

Then I think `run` should get a `--stream` flag or something. Keeping two different non-interactive code paths is just inviting more bugs like #139. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/140)*

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[main.go, line 857 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/140#-L8mikwyQvaHKUGtbzf6-r1-857:-L8mxele2Fi0Bkn-Kr_o:b-8a8ltz) ([raw file](https://github.com/cockroachdb/roachprod/blob/b8cd2ca2269720ec8dbbd6984cb4f9a12c6492ab/main.go#L857)):*

There would still be 2 code paths, we'd just be moving the switch point from here to `Run`. That would probably have eliminated #139. Another option is to merge `Cluster.Ssh` into `Cluster.Run`. Let me see how that looks.
</blockquote></details>

Reworked this so that `Run` streams if the command is only being run on a single node. This also subsumes #139. PTAL.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/140)*

PR #141, opened 2018-03-29
Allow overriding --store and --locality when starting a cluster

### LINKED ISSUE #131: Allow overriding flags passed to start
`roachprod start` does not currently allow override the flags passed to `cockroach start`. This prevents specifying a different `--store` directory (you can add a second store, but not change the value of the default one) or overriding the value of the `--locality` flag. One possibility is that `roachprod start` can snoop on the `--args` flag and omit any flag that is specified there.

Cc @tschottdorf

petermattis:
Snoop on the contents `--args`, looking for `--store` or `--locality`
flags. If specified, `start` will not automatically provide those
flags. This is necessary because both flags append to lists instead of
overwriting a single value.

Fixes #131

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/141)

### DIFF
--- a/install/cockroach.go
+++ b/install/cockroach.go
@@ -89,6 +89,15 @@ func GetAdminUIPort(connPort int) int {
 	return connPort + 1
 }
 
+func argExists(args []string, target string) bool {
+	for _, arg := range args {
+		if arg == target || strings.HasPrefix(arg, target+"=") {
+			return true
+		}
+	}
+	return false
+}
+
 func (r Cockroach) Start(c *SyncedCluster, extraArgs []string) {
 	// Check to see if node 1 was started indicating the cluster was
 	// bootstrapped.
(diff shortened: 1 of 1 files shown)


benesch:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---

*[install/cockroach.go, line 298 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/141#-L8mqqmb2OYcuDmdwmfp:-L8mqqmb2OYcuDmdwmfq:bb2ihth) ([raw file](https://github.com/cockroachdb/roachprod/blob/9f207793c77a6a3617c121fbbae89322796dd0e4/install/cockroach.go#L298)):*
> ```Go
> 			logDir = fmt.Sprintf("${HOME}/local/%d/logs", nodes[i])
> 		}
> 		if !argExists(extraArgs, "--store=") {
> ```

I tend to use the form `--store FOO`. Does looking for just `--store` work?

---

*[install/cockroach.go, line 317 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/141#-L8mqux33h79xs6f35hD:-L8mqux33h79xs6f35hE:b-7esevx) ([raw file](https://github.com/cockroachdb/roachprod/blob/9f207793c77a6a3617c121fbbae89322796dd0e4/install/cockroach.go#L317)):*
> ```Go
> 		args = append(args, fmt.Sprintf("--http-port=%d", GetAdminUIPort(port)))
> 		if locality := c.locality(nodes[i]); locality != "" {
> 			if !argExists(extraArgs, "--locality=") {
> ```

Ditto here.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/141)*

tschottdorf:
LGTM, thanks!

PR #146, opened 2018-04-06
Remove the use of ssh for "local" clusters

### LINKED ISSUE #128: Make roachprod --local work without ssh
I tried to start running some of our roachtests on every PR using --local, but the builder image doesn't have ssh (and probably shouldn't). It's also an extra step for each dev to set up passwordless ssh to localhost before they can use --local

petermattis:
Add `session` interface and `{local,remote}Session` implementations. The
`localSession` is a small wrapper around `exec.Cmd` while the remote
session is a small wrapper around `ssh.Session`. Commands executed with
`localSession` are always run via `/bin/bash -c` which has the nice
side-effect of removing any dependency on the users shell.

Fixes #128

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/146)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -601,25 +601,88 @@ func (c *SyncedCluster) Get(src, dest string) {
 		wg.Add(1)
 		go func(i int) {
 			defer wg.Done()
-			session, err := ssh.NewSSHSession(c.user(c.Nodes[i]), c.host(c.Nodes[i]))
 
+			src := src
+			dest := dest
+			if len(c.Nodes) > 1 {
+				base := fmt.Sprintf("%d.%s", c.Nodes[i], filepath.Base(dest))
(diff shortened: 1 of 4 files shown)


petermattis:
I think I got all of the uses of local ssh. If I missed something, `ssh.NewSSHSession` will now return an error when trying to open a session to `localhost`.

petermattis:
@benesch Please take it for a quick spin and see if you notice any problems. Note that `roachtest run -l` currently doesn't work but is fixed by https://github.com/cockroachdb/cockroach/pull/24567 (feel to rubber stamp that PR).

benesch:
👍 I'm traveling so I might not be able to get to it until Monday.

petermattis:
@benesch Did you want to take this for a spin?

PR #147, opened 2018-04-06
Fix binary flag default value

garvitjuniwal:
roachprod looks for the binary name in PATH
first and then in GOPATH, but looking up in PATH
does not work if the name has a / in it. Fix the 
default value of the binary flag to not have a /.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/147)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,7 +6,7 @@ import (
 )
 
 var (
-	Binary     = "./cockroach"
+	Binary     = "cockroach"
 	SlackToken string
 	OSUser     *user.User
 )
--- a/main.go
+++ b/main.go
@@ -1182,7 +1182,7 @@ func main() {
 		switch cmd {
 		case startCmd, testCmd:
 			cmd.Flags().StringVarP(
-				&config.Binary, "binary", "b", "./cockroach",
+				&config.Binary, "binary", "b", config.Binary,
 				"the remote cockroach binary used to start a server")
 			cmd.Flags().BoolVar(
 				&install.StartOpts.Sequential, "sequential", false,


petermattis:
:lgtm:

FYI, the use of `./cockroach` was to allow finding the binary in the current directory. At least, it had that effect at one time. I'm not seeing how that would work with the current code. 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---

*[main.go, line 1185 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/147#-L9RkCgc1HyvKwPu2HpJ:-L9RkCgc1HyvKwPu2HpK:bsjaaeq) ([raw file](https://github.com/cockroachdb/roachprod/blob/5e92513681b4d6c8cb0825decbde40f1e814d671/main.go#L1185)):*
> ```Go
> 		case startCmd, testCmd:
> 			cmd.Flags().StringVarP(
> 				&config.Binary, "binary", "b", "cockroach",
> ```

`s/"cockroach"/config.Binary/g`

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/147#-:-L9RkGUFD3Qb88eyC83_:bb6i3ed)*

garvitjuniwal:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[main.go, line 1185 at r1](https://beta.reviewable.io/reviews/cockroachdb/roachprod/147#-L9RkCgc1HyvKwPu2HpJ:-L9Rrzh61VLeSIrBCwF-:b-896fix) ([raw file](https://github.com/cockroachdb/roachprod/blob/5e92513681b4d6c8cb0825decbde40f1e814d671/main.go#L1185)):*

`s/"cockroach"/config.Binary/g`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/roachprod/147)*

nvanbenschoten:
@bdarnell was suspicious that this was the cause of the last night's failures. I just dug through the artifacts and it looks like all tests (except Jepsen!) failed with the error:
```
0: Process exited with status 127
~ cockroach version
bash: cockroach: command not found
```
It looks like this confirms Ben's theory. 

@petermattis you said above: 
> I'm not seeing how that would work with the current code

What changed?

petermattis:
> What changed?

Teamcity was probably relying on the fact that `./cockroach` was in the current working directory. Let me take a look.

petermattis:
Actually, it looks like `cockroachNodeBinary` is being used inappropriately. Let me track this down.

PR #148, opened 2018-04-10
Fix specification of remote cockroach binary

petermattis:
When running `cockroach` remotely, ensure that path is relative to the
current directory as `PATH` is not set up to contain the current
directory (i.e. the home directory).

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/148)


petermattis:
Adjusted the code so that I could test this locally. Definitely fixes the problem.

petermattis:
Also verified that `roachtest run kv95/nodes=1` works.

PR #149, opened 2018-04-16
Fix `local {sql,ssh}` to not use ssh

petermattis:
Fix the `local {sql,ssh}` commands to not use ssh.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/149)


petermattis:
@benesch Not terribly important, but do you have time to give this a rubber stamp?

PR #153, opened 2018-04-19
bug: Fixes `roachprod ssh ...` not returning error when remote fails.

### LINKED ISSUE #152: `roachprod ssh ...` not returning error when remote's exit status is not 0
When firing the `roachprod ssh` command, it doesn't return an error even though the remote 
returns an error. This is causing `func RunE` in `cockroach/pkg/cmd/roachtest/cluster.go`
to not return an error even though the command fails.

windchan7:
Fixes the issue that when running a command on a node thorugh `roachprod ssh`,
it will return the remote error appropriately.

Fixes #152.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/153)

### DIFF
--- a/main.go
+++ b/main.go
@@ -865,8 +865,7 @@ var runCmd = &cobra.Command{
 		if len(title) > 30 {
 			title = title[:27] + "..."
 		}
-		_ = c.Run(os.Stdout, c.Nodes, title, cmd)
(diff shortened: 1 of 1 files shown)


windchan7:
Thanks!

PR #155, opened 2018-04-23
Separate stdout/stderr `run` is invoked in streaming mode

### LINKED ISSUE #154: Maintain stdout/stderr separation
`roachprod run` combines both stdout and stderr of its command into roachprod's stderr. This breaks the jepsen test's [log collection](https://github.com/cockroachdb/cockroach/blob/9da732ca6a06c12149213c97fd8215ca23067eba/pkg/cmd/roachtest/jepsen.go#L226). When running a single command, both stdout and stderr should be passed to their corresponding file descriptors in the roachprod process.

petermattis:
When `run` is invoked on a single target node (i.e. streaming mode),
separate stdout and stderr like a good citizen.

Fixes #154

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/155)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -293,7 +293,7 @@ done
 	return ch
 }
 
-func (c *SyncedCluster) Run(w io.Writer, nodes []int, title, cmd string) error {
+func (c *SyncedCluster) Run(stdout, stderr io.Writer, nodes []int, title, cmd string) error {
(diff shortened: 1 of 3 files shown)


petermattis:
There was no reason for the previous behavior, it happened by accident when adding support for streaming `{ssh,run}` results was added.

PR #161, opened 2018-05-23
Add --treedist flag for put (defaults to true)

petermattis:
The treedist algorithm puts a file to multiple nodes in a cluster by first
copying the file to a single node, then using that node as a source for
"fanout" (2) additional copies. Then using the result 1+fanout nodes as
sources for additional copies, etc.

The treedist algorithm can copy the 113MB cockroach binary to a 10 node
cluster 40% faster than the non-treedist approach from our corp network. On
a slower network the speedup should be even more dramatic.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/161)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -51,7 +51,6 @@ type SyncedCluster struct {
 	Args    []string
 	Tag     string
 	Impl    ClusterImpl
-	UseSCP  bool
 }
 
 func (c *SyncedCluster) host(index int) string {
@@ -479,10 +478,12 @@ func formatProgress(p float64) string {
 }
 
 func (c *SyncedCluster) Put(src, dest string) {
-	// TODO(peter): Only put 10 nodes at a time. When a node completes, output a
-	// line indicating that.
+	// TODO(peter): Do something akin to treedist. Copy src to first node. Then
+	// copy from first node to 2 nodes. Repeat. Keep 2 outstanding copies from
+	// each source. Only use the local node for the initial copy.
+
 	var detail string
-	if !c.IsLocal() && c.UseSCP {
+	if !c.IsLocal() {
 		detail = " (scp)"
 	}
 	fmt.Printf("%s: putting%s %s %s\n", c.Name, detail, src, dest)
@@ -521,24 +522,11 @@ func (c *SyncedCluster) Put(src, dest string) {
 				return
 			}
 
-			if c.UseSCP {
-				to := dest
-				if c.IsLocal() {
-					to = fmt.Sprintf(os.ExpandEnv("${HOME}/local/%d/%s"), c.Nodes[i], dest)
-				}
-				err := c.scp(src, fmt.Sprintf("%s@%s:%s", c.user(c.Nodes[i]), c.host(c.Nodes[i]), to))
-				results <- result{i, err}
+			to := dest
+			if c.IsLocal() {
+				to = fmt.Sprintf(os.ExpandEnv("${HOME}/local/%d/%s"), c.Nodes[i], dest)
 			}
-
-			session, err := ssh.NewSSHSession(c.user(c.Nodes[i]), c.host(c.Nodes[i]))
-			if err == nil {
-				defer session.Close()
-				err = ssh.SCPPut(src, dest, func(p float64) {
-					linesMu.Lock()
-					defer linesMu.Unlock()
-					lines[i] = formatProgress(p)
-				}, session)
-			}
+			err := c.scp(src, fmt.Sprintf("%s@%s:%s", c.user(c.Nodes[i]), c.host(c.Nodes[i]), to))
 			results <- result{i, err}
 		}(i)
 	}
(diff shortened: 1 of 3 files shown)


petermattis:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[install/cluster_synced.go, line 911 at r2](https://reviewable.io/reviews/cockroachdb/roachprod/161#-LDCpuuOBEz1THc_DgFQ:-LDCukzl4V9B-QvpIVlO:b-z8dg72) ([raw file](https://github.com/cockroachdb/roachprod/blob/f555bd8397eff9be9f8abe3e08b5eca69867113e/install/cluster_synced.go#L911)):*

The need for ForwardAgent is unfortunate. It's a bit of a security weakness since roachprod workers aren't exactly hardened. 

For a non-ForwardAgent alternative, we may want to move [the jepsen test's ssh setup](https://github.com/cockroachdb/cockroach/blob/10416c40ad6758dbfbe6188b6d659895ecc32def/pkg/cmd/roachtest/jepsen.go#L115-L133) into `roachprod create` so that roachprod workers can ssh to each other without having an agent forwarded in from the outside.

Or we could use a protocol other than scp. It would be pretty easy to start an unauthenticated rsync daemon on the source nodes and pull from there. Or there are other tools I'm not personally familiar with (twitter has something bittorrent-based, and `udpcast` came up in a quick search).
</blockquote></details>

How bad is the security weakness? I'm assuming the problem is anyone able to access a roachprod node can then get access to the agent's identity which will give them access to other production nodes. But if they can already access a node how bad is that? I'm probably missing something.

Using an unauthenticated protocol feels like more of a risk than a forwarded agent. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/161)*

bdarnell:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[install/cluster_synced.go, line 911 at r2](https://reviewable.io/reviews/cockroachdb/roachprod/161#-LDCpuuOBEz1THc_DgFQ:-LDCvh_EARShDPlwjDWd:bolocnw) ([raw file](https://github.com/cockroachdb/roachprod/blob/f555bd8397eff9be9f8abe3e08b5eca69867113e/install/cluster_synced.go#L911)):*

How bad is the security weakness? I'm assuming the problem is anyone able to access a roachprod node can then get access to the agent's identity which will give them access to other production nodes. But if they can already access a node how bad is that? I'm probably missing something.

Using an unauthenticated protocol feels like more of a risk than a forwarded agent. 
</blockquote></details>

It gets you access to all the other keys stored in the same agent. So, for example, assuming you have the ssh key that you use for github loaded into your ssh agent while running roachprod, I could access github repos *as you* via a roachprod worker. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/161)*

bdarnell:
Also, what about multi-region clusters? It's not crucial, but making the fanout aware of the network topology would be nice.

petermattis:
> Also, what about multi-region clusters? It's not crucial, but making the fanout aware of the network topology would be nice.

Yeah, this doesn't take topology into account. I'll add a TODO. 

---

Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[install/cluster_synced.go, line 911 at r2](https://reviewable.io/reviews/cockroachdb/roachprod/161#-LDCpuuOBEz1THc_DgFQ:-LDDMlc87iQEC5W1Yy_Z:bjb66gw) ([raw file](https://github.com/cockroachdb/roachprod/blob/f555bd8397eff9be9f8abe3e08b5eca69867113e/install/cluster_synced.go#L911)):*

It gets you access to all the other keys stored in the same agent. So, for example, assuming you have the ssh key that you use for github loaded into your ssh agent while running roachprod, I could access github repos *as you* via a roachprod worker. 
</blockquote></details>

Is this concerning enough not to merge as-is? I'm not quite sure what your LGTM meant.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/161#-:-LDDMiQs190lAleLlXLZ:biq1nbe)*

bdarnell:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[install/cluster_synced.go, line 911 at r2](https://reviewable.io/reviews/cockroachdb/roachprod/161#-LDCpuuOBEz1THc_DgFQ:-LDDOfDT-2RneLmHUBtD:b-9qcppy) ([raw file](https://github.com/cockroachdb/roachprod/blob/f555bd8397eff9be9f8abe3e08b5eca69867113e/install/cluster_synced.go#L911)):*

Is this concerning enough not to merge as-is? I'm not quite sure what your LGTM meant.
</blockquote></details>

I'm OK merging it as-is. My preferred next-step solution would be to set up ssh keys between nodes like we do in jepsen instead of switching to a completely different tool.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/161)*

PR #164, opened 2018-05-24
Add --json flag to roachprod list command

nvanbenschoten:
This flag sets the format of the command output to json. This will
make the output more easily ingestible by other tools.

The PR also makes two minor changes to the VM struct:
- it adds a MachineType field
- it strips the URL path from the VPC string for GCE VMs

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/164)

### DIFF
--- a/cloud/cluster_cloud.go
+++ b/cloud/cluster_cloud.go
@@ -15,10 +15,10 @@ import (
 const vmNameFormat = "user-<clusterid>-<nodeid>"
 
 type Cloud struct {
-	Clusters map[string]*CloudCluster
+	Clusters map[string]*CloudCluster `json:"clusters"`
 	// Any VM in this list can be expected to have at least one element
 	// in its Errors field.
-	BadInstances vm.List
+	BadInstances vm.List `json:"bad_instances"`
 }
 
 // Collate Cloud.BadInstances by errors.
@@ -50,12 +50,12 @@ func newCloud() *Cloud {
 //
 // TODO(benesch): unify with syncedCluster.
 type CloudCluster struct {
-	Name string
-	User string
+	Name string `json:"name"`
+	User string `json:"user"`
 	// This is the earliest creation and shortest lifetime across VMs.
-	CreatedAt time.Time
-	Lifetime  time.Duration
-	VMs       vm.List
+	CreatedAt time.Time     `json:"created_at"`
+	Lifetime  time.Duration `json:"lifetime"`
+	VMs       vm.List       `json:"vms"`
 }
 
 // Clouds returns the names of all of the various cloud providers used
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"os"
@@ -57,6 +58,7 @@ var (
 	dryrun         bool
 	extendLifetime time.Duration
 	listDetails    bool
+	listJSON       bool
 	listMine       bool
 	clusterType    = "cockroach"
 	secure         = false
@@ -324,7 +326,7 @@ Local Clusters
 					}
 				}
 
-				if err := syncAll(cloud); err != nil {
+				if err := syncAll(cloud, false /* quiet */); err != nil {
 					return err
 				}
 			}
@@ -453,7 +455,7 @@ The third and fourth columns are the number of nodes in the cluster and the
 time remaining before the cluster will be automatically destroyed. Note that
 local clusters do not have an expiration.
 
-The --details adjusts the output format to include per-node details:
+The --details flag adjusts the output format to include per-node details:
 
   ~ roachprod list --details
   local [local]: (no expiration)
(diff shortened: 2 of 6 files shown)


petermattis:
:lgtm:

---

Reviewed 4 of 4 files at r1, 3 of 3 files at r2.
Review status: 5 of 6 files reviewed at latest revision, all discussions resolved.

---

*[main.go, line 543 at r2](https://reviewable.io/reviews/cockroachdb/roachprod/164#-LDEa7QZ2dvpBXD-Pu0K:-LDEa7QZ2dvpBXD-Pu0L:b-c8vhpb) ([raw file](https://github.com/cockroachdb/roachprod/blob/ca3b72fcd22a6caa9aa6b103b1efe480d0bad1ff/main.go#L543)):*
> ```Go
> 
> 			enc := json.NewEncoder(os.Stdout)
> 			enc.SetIndent("", "  ")
> ```

If I'm only encoding a single json value, I usually use `json.Marshal` or `json.MarshalIndent`. I don't think there is any need for a `json.Encoder` here.

---

*[vm/gce/gcloud.go, line 106 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/164#-LDE_hXX0hdZIDhjBIqD:-LDE_hXX0hdZIDhjBIqE:b-xb4jq2) ([raw file](https://github.com/cockroachdb/roachprod/blob/6871cfa5db10bf7754161d611e9b663179d07e6d/vm/gce/gcloud.go#L106)):*
> ```Go
> 		return s[len(s)-1]
> 	}
> 	machineType := lastComponent(jsonVM.MachineType)
> ```

What is this about? I thought GCE machine types had hyphens, not slashes. For example, `n1-standard-4`.  Perhaps provide an example in a comment of what a typical `jsonVM.MachineType` looks like.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/164#-:-LDEaS-dBUWw9q3yvxEH:bnfp4nl)*

nvanbenschoten:
Review status: 5 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[main.go, line 543 at r2](https://reviewable.io/reviews/cockroachdb/roachprod/164#-LDEa7QZ2dvpBXD-Pu0K:-LDFAHyo3q8U8rLCadz0:bf9dz7s) ([raw file](https://github.com/cockroachdb/roachprod/blob/ca3b72fcd22a6caa9aa6b103b1efe480d0bad1ff/main.go#L543)):*

If I'm only encoding a single json value, I usually use `json.Marshal` or `json.MarshalIndent`. I don't think there is any need for a `json.Encoder` here.
</blockquote></details>

Not that it matters much here, but doesn't that result in extra an data copy because you're no longer encoding directly into the `io.Writer`?

---

*[vm/gce/gcloud.go, line 106 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/164#-LDE_hXX0hdZIDhjBIqD:-LDFAo4p70p-vaqMV68s:bqghelw) ([raw file](https://github.com/cockroachdb/roachprod/blob/6871cfa5db10bf7754161d611e9b663179d07e6d/vm/gce/gcloud.go#L106)):*

What is this about? I thought GCE machine types had hyphens, not slashes. For example, `n1-standard-4`.  Perhaps provide an example in a comment of what a typical `jsonVM.MachineType` looks like.
</blockquote></details>

`jsonVM.MachineType` ends up as a URL that looks like `"https://www.googleapis.com/compute/v1/projects/cockroach-nathan/zones/us-east1-b/machineTypes/n1-standard-16"`. This is the same issue we run into with Zone and VPC.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/164)*

nvanbenschoten:
Review status: 5 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[main.go, line 543 at r2](https://reviewable.io/reviews/cockroachdb/roachprod/164#-LDEa7QZ2dvpBXD-Pu0K:-LDHlsSYBpDZVDunnWJ-:b3a9mir) ([raw file](https://github.com/cockroachdb/roachprod/blob/ca3b72fcd22a6caa9aa6b103b1efe480d0bad1ff/main.go#L543)):*

Yeah, there would be an extra copy, but it definitely doesn't matter here. 
</blockquote></details>

I'm just going to keep this as-is because the other approach requires 2 error checks and an extra line of code.

---

*[vm/gce/gcloud.go, line 106 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/164#-LDE_hXX0hdZIDhjBIqD:-LDHlrEk8uFI3Ded1tvL:b-f8sf1p) ([raw file](https://github.com/cockroachdb/roachprod/blob/6871cfa5db10bf7754161d611e9b663179d07e6d/vm/gce/gcloud.go#L106)):*

Ack. Thanks.
</blockquote></details>

Added a comment.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/164)*

Issue #163, opened 2018-05-23
Avoid ForwardAgent with --treedist


petermattis:
In order to perform a remote-to-remote copy, `put --treedist` needs to forward the agent. Rather than performing agent forwarding we could instead distribute keys to each of the remote nodes to allow internode ssh. For example, the jepsen roachtests do the following at setup:

```go
	// SSH setup: create a key on the controller.
	tempDir, err := ioutil.TempDir("", "jepsen")
	if err != nil {
		t.Fatal(err)
	}
	c.Run(ctx, controller, "sh", "-c", `"test -f .ssh/id_rsa || ssh-keygen -f .ssh/id_rsa -t rsa -N ''"`)
	pubSSHKey := filepath.Join(tempDir, "id_rsa.pub")
	cmd := c.LoggedCommand(ctx, roachprod, "get", c.makeNodes(controller), ".ssh/id_rsa.pub", pubSSHKey)
	if err := cmd.Run(); err != nil {
		t.Fatal(err)
	}
	// TODO(bdarnell): make this idempotent instead of filling up .ssh configs.
	c.Put(ctx, pubSSHKey, "controller_id_rsa.pub", workers)
	c.Run(ctx, workers, "sh", "-c", `"cat controller_id_rsa.pub >> .ssh/authorized_keys"`)
	// Prime the known hosts file, and use the unhashed format to
	// work around JSCH auth error: https://github.com/jepsen-io/jepsen/blob/master/README.md
	for _, ip := range c.InternalIP(ctx, workers) {
		c.Run(ctx, controller, "sh", "-c", fmt.Sprintf(`"ssh-keyscan -t rsa %s >> .ssh/known_hosts"`, ip))
	}
```

PR #165, opened 2018-05-24
Default --treedist to false

petermattis:
It currently requires a local ssh agent to be running for
authentication, which isn't true when run via the roachtest
nightlies. #163 will remove the agent requirement.

See #163

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/165)

### DIFF
--- a/main.go
+++ b/main.go
@@ -65,7 +65,7 @@ var (
 	tag            string
 	external       = false
 	adminurlOpen   = false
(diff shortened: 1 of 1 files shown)


nvanbenschoten:
Is this only a problem for nightly roachtests or do you expect it to cause issues for manual roachtests as well?

PR #168, opened 2018-05-24
Setup ssh keys at cluster creation

### LINKED ISSUE #163: Avoid ForwardAgent with --treedist

petermattis:
When a cluster is created, generate an ssh key and distribute it to all
of the nodes in the cluster allowing easy intra-cluster ssh. Also
populate the `.ssh/known_hosts` with internal and external IP
addresses (again to ease intra-cluster ssh).

Fixes #163

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/168)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -2,6 +2,7 @@ package install
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


petermattis:
I've verified that `put --treedist` works without requiring `ForwardAgent` or any of the previous fanciness with using `StrictHostKeyChecking=no` between remote nodes.

bdarnell:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved.

---

*[install/cluster_synced.go, line 408 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/168#-LDZUT-O9PNrA4bKF3pF:-LDZUT-P0pUFumrqpnAK:b3raubs) ([raw file](https://github.com/cockroachdb/roachprod/blob/6ac31a97fb700fb13b58459a522372ac95b8897b/install/cluster_synced.go#L408)):*
> ```Go
> 	var msg string
> 	var sshTar []byte
> 	c.Parallel("generating ssh key", 1, 0, func(i int) ([]byte, error) {
> ```

`c.Parallel` for a single node?

---

*[install/cluster_synced.go, line 424 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/168#-LDZWK5X-frzPlkGEXQu:-LDZWK5X-frzPlkGEXQv:bddypzc) ([raw file](https://github.com/cockroachdb/roachprod/blob/6ac31a97fb700fb13b58459a522372ac95b8897b/install/cluster_synced.go#L424)):*
> ```Go
> tar cf - .ssh/id_rsa .ssh/id_rsa.pub .ssh/authorized_keys
> `
> 		sshTar, err = session.CombinedOutput(cmd)
> ```

This should be `Output`, not `CombinedOutput`. `stderr` should be kept separate from the tar data. 

---

*[install/cluster_synced.go, line 454 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/168#-LDZWkAcC5GZ6p7h4t3E:-LDZWkAcC5GZ6p7h4t3F:b-cmfde) ([raw file](https://github.com/cockroachdb/roachprod/blob/6ac31a97fb700fb13b58459a522372ac95b8897b/install/cluster_synced.go#L454)):*
> ```Go
> 
> 	// Populate the known_hosts file with both internal and external IPs of all
> 	// of the nodes in the cluster.
> ```

Comment that as a side effect, this creates the known hosts file in unhashed format, working around a limitation of jsch (which is used in jepsen tests). 

---

*[install/cluster_synced.go, line 471 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/168#-LDZWue_CbONp9tV5jCM:-LDZWue_CbONp9tV5jCN:b-f3zvzw) ([raw file](https://github.com/cockroachdb/roachprod/blob/6ac31a97fb700fb13b58459a522372ac95b8897b/install/cluster_synced.go#L471)):*
> ```Go
> 		defer session.Close()
> 
> 		cmd := `ssh-keyscan -t rsa ` + strings.Join(ips, " ") + ` >> .ssh/known_hosts`
> ```

This is not idempotent and will result in the known hosts file growing over time if this is repeated. (the current jepsen tests have the same issue, though). 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/168#-:-LDZXPQ3CvR8GIykO5H9:bnfp4nl)*

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[install/cluster_synced.go, line 408 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/168#-LDZUT-O9PNrA4bKF3pF:-LDb1fQNAGj9lrObF08g:bndfihu) ([raw file](https://github.com/cockroachdb/roachprod/blob/6ac31a97fb700fb13b58459a522372ac95b8897b/install/cluster_synced.go#L408)):*

`c.Parallel` for a single node?
</blockquote></details>

Yeah, I do this in other places too because it prints the message and progress info.

---

*[install/cluster_synced.go, line 424 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/168#-LDZWK5X-frzPlkGEXQu:-LDb2Tov7DD6zGqUelrI:b-lkemsa) ([raw file](https://github.com/cockroachdb/roachprod/blob/6ac31a97fb700fb13b58459a522372ac95b8897b/install/cluster_synced.go#L424)):*

This should be `Output`, not `CombinedOutput`. `stderr` should be kept separate from the tar data. 
</blockquote></details>

Our local `session` interface doesn't expose `Output`, and even if it didn't I don't see how to get stderr if you use that method. I've changed this to put stdout and stderr into separate `byte.Buffers` this seems to be what `Output` and `CombinedOutput` use internally. 

---

*[install/cluster_synced.go, line 454 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/168#-LDZWkAcC5GZ6p7h4t3E:-LDb2p0F-5p4OtQzSTZh:b-896fix) ([raw file](https://github.com/cockroachdb/roachprod/blob/6ac31a97fb700fb13b58459a522372ac95b8897b/install/cluster_synced.go#L454)):*

Comment that as a side effect, this creates the known hosts file in unhashed format, working around a limitation of jsch (which is used in jepsen tests). 
</blockquote></details>

Done.

---

*[install/cluster_synced.go, line 471 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/168#-LDZWue_CbONp9tV5jCM:-LDb2qQ8CHmbfLYpcQra:baft2gk) ([raw file](https://github.com/cockroachdb/roachprod/blob/6ac31a97fb700fb13b58459a522372ac95b8897b/install/cluster_synced.go#L471)):*

This is not idempotent and will result in the known hosts file growing over time if this is repeated. (the current jepsen tests have the same issue, though). 
</blockquote></details>

Is that an issue here given that we only run this once per cluster?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/168)*

PR #181, opened 2018-07-18
Rework the mechanism to wait for a node to be initialized

petermattis:
The startup script now touches /mnt/data1/.roachprod-initialized when
initialization is coimplete. SyncedCluster.Wait waits for that file to
appear. The hope is that this is more robust than waiting for systemctl
and it also works on AWS.


petermattis:
> It occurs to me that one downside to this approach is that it's not robust to restarts. Good enough for roachtests, though.

We don't stop roachprod nodes. Do we?

PR #183, opened 2018-07-29
Atomically write the hosts files when syncing hosts

petermattis:
The file lock acquired by syncAll prevents another process from syncing
the hosts file concurrently, but we don't acquire that lock while
reading the hosts files. This could lead to a concurrent reader seeing
an empty hosts file.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/183)

### DIFF
--- a/hosts.go
+++ b/hosts.go
@@ -28,24 +28,39 @@ func initHostDir() error {
 func syncHosts(cloud *cloud.Cloud) error {
 	hd := os.ExpandEnv(config.DefaultHostDir)
 
-	// Write all host files.
+	// Write all host files. We're the only process doing this due to the file
+	// lock acquired by syncAll, but other processes may be reading the host
+	// files concurrently so we need to write the files atomically by writing to
+	// a temporary file and renaming.
 	for _, c := range cloud.Clusters {
 		filename := path.Join(hd, c.Name)
-		file, err := os.Create(filename)
-		if err != nil {
-			return errors.Wrapf(err, "problem creating file %s", filename)
-		}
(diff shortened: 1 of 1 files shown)


benesch:
Oops. Good catch. 

I'm having trouble convincing myself this is correct, though. `loadClusters` reads all files in the hosts directory, so it's going to pick up these `.tmp` files. Maybe that's fine? But I think it might be cleaner to just acquire the flock in `loadClusters` instead of making each file write atomic.

petermattis:
> I'm having trouble convincing myself this is correct, though. loadClusters reads all files in the hosts directory, so it's going to pick up these .tmp files. Maybe that's fine? But I think it might be cleaner to just acquire the flock in loadClusters instead of making each file write atomic.

`syncAll` is somewhat slow, though. I don't think it is a big problem that `loadClusters` sees the `.tmp` files. It will just create clusters with those file names. But also easy to skip those files.

benesch:
SGTM

petermattis:
I should mention for posterity that I discovered this issue because I'm running the `kv{0,95}` roachtests in a loop, back filling performance numbers for binaries over the course of the 2.1 release cycle. Before this change I would very quickly hit an error when creating a cluster where it appeared the cluster had 0 nodes. The relevant factor here was that I was running `roachtest run 'kv(0|95)/encrypt=false'`. This runs 4 tests concurrently and thus creates 4 clusters concurrently. After this change, I've successfully run 5 back fill iterations without incident.

PR #184, opened 2018-07-30
Default --treedist to false

petermattis:
A handful of nightly tests failed with errors that look related to
treedist (`roachprod put` failures). Turning it off by default until I
have time to investigate.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/184)

### DIFF
--- a/main.go
+++ b/main.go
@@ -68,7 +68,7 @@ var (
 	tag            string
 	external       = false
 	adminurlOpen   = false
-	useTreeDist    = true
+	useTreeDist    = false
 	encrypt        = false
 	quiet          = false
 	sig            = 9


tschottdorf:
LGTM

petermattis:
FYI, I've been running a backfill of performance numbers using `roachtest run 'kv(0|95)`' this weekend with a roachprod with treedist enabled and have not seen a single failure of `roachprod put`. That makes the failures with the nightly tests quite a bit more surprising.

a-robinson:
I've always had big problems with "Host key verification failed" errors when using `--treedist` on both big and small clusters and even when I completely clear out my `known_hosts` file.

petermattis:
@a-robinson The host key verification is taking place on the cluster nodes, so your local `.ssh/known_hosts` has no effect.

PR #187, opened 2018-08-10
Increase SQL connection timeout delay

nvanbenschoten:
See https://github.com/cockroachdb/cockroach/issues/28369.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/187)

### DIFF
--- a/install/cockroach.go
+++ b/install/cockroach.go
@@ -387,6 +387,7 @@ tar cvf certs.tar certs
 			if c.IsLocal() {
 				cmd = `cd ${HOME}/local/1 ; `
 			}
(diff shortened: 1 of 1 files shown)


benesch:
Roachprod always uses Bash, so we omit the /usr/bin/env everywhere.

knz:
acknowledged, thanks for rexplaining

knz:
Consider setting the env var to "0" to disable the timeout. This will make the client wait forever.

nvanbenschoten:
> Consider setting the env var to "0" to disable the timeout. This will make the client wait forever.

Done.

PR #189, opened 2018-08-29
roachprod: disambiguate confusing log

tschottdorf:
`roachprod get` is called by `roachtest` when collecting logs, but the
output "failed" isn't linked to that operation in particular and looks
like something funky might be going on in the test. Be more specific.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/189)


tschottdorf:
TFTR!

PR #194, opened 2018-09-01
Only initialize cluster settings once

petermattis:
Previously we were initializing cluster settings every time node 1
restarted. This wasn't harmful, but it was a bit slow (a few seconds)
and noticeable in some recently added roachtests.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/194)

### DIFF
--- a/install/cockroach.go
+++ b/install/cockroach.go
@@ -387,13 +387,18 @@ tar cvf certs.tar certs
 			if c.IsLocal() {
 				cmd = `cd ${HOME}/local/1 ; `
 			}
-			cmd += `COCKROACH_CONNECT_TIMEOUT=0 `
-			cmd += cockroachNodeBinary(c, 1) + " sql --url " +
+			dir := c.Impl.NodeDir(c, nodes[i])
+			cmd += `
+if ! test -e ` + dir + `/settings-initialized ; then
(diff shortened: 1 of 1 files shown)


install/cockroach.go
```diff
@@ -387,13 +387,18 @@ tar cvf certs.tar certs
...
+			dir := c.Impl.NodeDir(c, nodes[i])
+			cmd += `
+if ! test -e ` + dir + `/settings-initialized ; then
+  COCKROACH_CONNECT_TIMEOUT=0 ` + cockroachNodeBinary(c, 1) + " sql --url " +
 				r.NodeURL(c, "localhost", r.NodePort(c, 1)) + " -e " +
 				fmt.Sprintf(`"
 SET CLUSTER SETTING server.remote_debugging.mode = 'any';
 SET CLUSTER SETTING cluster.organization = 'Cockroach Labs - Production Testing';
-SET CLUSTER SETTING enterprise.license = '%s';"`, license)
+SET CLUSTER SETTING enterprise.license = '%s';"`, license) + `;
```
benesch: Do we set `set -e` when we invoke Bash? Otherwise all of these SET CLUSTER SETTINGS could fail but we’d still write the marker file.

petermattis: Good point. Switched to using `&&` instead of `;` to separate the `cockroach sql` command and the `touch` command.

PR #195, opened 2018-09-02
Add a restore-prebuilt-data-dir command

andreimatei:
The new command restores saved data dirs onto a new cluster. The dirs
are supposed to come from another cluster where a desired data set has
been loaded.
This speeds-up the "restore" of the TPCC-1000 fixture from ~40m to ~2m
on a 4 node cluster.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/195)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -1163,3 +1163,32 @@ func (c *SyncedCluster) Parallel(display string, count, concurrency int, fn func
 func (c *SyncedCluster) escapedTag() string {
 	return strings.Replace(c.Tag, "/", "\\/", -1)
 }
+
+// gs://cockroach-fixtures/workload/tpcc-prebuilt/
+func (c *SyncedCluster) RestorePrebuiltDataDir(path string) {
+	display := fmt.Sprintf("%s: restore prebuilt data dir", c.Name)
+	results := make([]string, len(c.Nodes))
+	c.Parallel(display, len(c.Nodes), 0, func(i int) ([]byte, error) {
+		session, err := c.newSession(c.Nodes[i])
+		if err != nil {
+			results[i] = err.Error()
+			return nil, nil
+		}
+		defer session.Close()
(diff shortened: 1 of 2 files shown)


andreimatei:
I've tried to play around to speed up creating a roachprod cluster with a TPCC fixture.
This patch doesn't work at the moment, and I don't know why - but hopefully you do!
What it does it restores data dirs saved from a different cluster with the same number of nodes. Is this supposed to work?
What I'm seeing is that, if I do `roachprod start` after the data dir have been restored, the nodes come up fine, except they don't contain the `tpcc` database that they're supposed to. However, the used capacity is listed as `219GB` - which I guess shows that the nodes are using the correct directory for the data, but nothing more? 
When starting node 1, I see this message in the log:
`initial startup completed, will now wait for `cockroach init` or a join to a running cluster to start accepting clients.`, but I don't know if it's relevant.
If anyone has a clue, or knows that this is not supposed to work...

andreimatei:
Never mind, I had screwed up.
I'll polish and ask for a review.

petermattis:
This is almost the same idea as `roachtest store-gen`, though that functionality is focused on generating the store directory dumps. For manual testing it probably makes sense to have the inverse of `store-gen` (`store-restore`?).

benesch:
How about `store-fetch`? We've tried to avoid "restore" to reduce confusion with actual backup/restore.

petermattis:
> How about store-fetch?

That naming works for me. Does this belong in `roachtest` given that it knows about the layout and location of store-gen directories.

andreimatei:
note to self: roachprod has moved to the main repo.

kenliu:
@andreimatei I'd like to close out this open PR since the code has been moved to the main repo. What do you need to do to close it?

andreimatei:
closing since there's been other work in the area

PR #197, opened 2018-09-05
fix bug in `stage` in which println was used when printf was intendend

vilterp:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/197)


petermattis:
LGTM

PR #199, opened 2018-09-06
Retry dialing tcp connections

petermattis:
Attempted workaround for cockroachdb/cockroach#29578. Will
try something more drastic if the problem persists.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/199)


petermattis:
PS The more drastic approach is to always fork/exec ssh. Might not be that difficult to do that.

benesch:
OK, then I blame the whole Go networking stack :)

PR #203, opened 2018-09-18
add uiport expander

tschottdorf:
This is going to be "necessary" (we could also hard-code the ports, but
that sounds wrong) for an improved decommissioning acceptance test which
verifies the /health?ready=1 endpoint.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/203)


tschottdorf:
Will do, added to my personal TODO list.

PR #204, opened 2018-09-19
Avoid looping on netcat in monitor

nvanbenschoten:
This fixes https://github.com/cockroachdb/cockroach/issues/27815. I'm not sure of a better way to detect when the parent process dies and I wasn't able to get an approach that tested stdout without writing to it (e.g. replacing `while :; do` with `while [ -t 1 ]; do`).

We could also push this logic into roachprod itself by catching the signal and killing all child processes before exiting. That's less clean though, and I'm afraid of any fallout that will have.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/204)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -261,6 +261,7 @@ while :; do
 
   if [ -n "${lastpid}" ]; then
     nc localhost %[1]d >/dev/null 2>&1
+    echo nc exited
   else
     sleep 1
   fi
@@ -275,6 +276,9 @@ done
 				return
 			}
 			// Give the session a valid stdin pipe so that nc won't exit immediately.
+			// When nc does exit, we write to stdout, which has a side effect of
+			// checking whether the stdout pipe has broken. This allows us to detect
+			// when the roachprod process is killed.
 			inPipe, err := session.StdinPipe()
 			if err != nil {
 				ch <- nodeMonitorInfo{nodes[i], err.Error()}


nvanbenschoten:
> but the recent change from using golang's ssh to forking ssh likely broke that fix.

That would explain why I was unable to reproduce cockroachdb/cockroach#27815 last week. I must have updated `roachprod` recently.

I'll wait on @benesch in case he has any magical incantations that can clean this up.

benesch:
This monitor stuff is the gift that keeps on giving. I don't have anything better to suggest. Here's an explanation of what's going on, though, if you're curious: https://serverfault.com/a/463375

> I'm pretty sure I fixed this before, but the recent change from using golang's ssh to forking ssh likely broke that fix. I don't recall the precise details about what my fix was, but it was something in regards to RequestPty.

Yep, agreed. I think when the SSH session lived in the roachprod process it was cleanly shutdown, which tickled the TCP session in a way that caused a SIGHUP or something to get propagated. Now that `exec.CommandContext` is `kill -9`ing the `ssh` binary (`exec.CommandContext` is a misfeature IMO, but I won't get into that here) the TCP session is just left hanging, which is apparently a perfectly valid state for a TCP session to be in. I think turning on keep-alives also works around this problem, but I think the solution that @nvanbenschoten merged is by far the most straightforward.

PR #207, opened 2018-09-27
Increase the default maximum number of open file descriptors

nvanbenschoten:
Fixes https://github.com/cockroachdb/cockroach/issues/30669.

Once this is merged, we can go back and remove any `ulimit -n`
invocations from roachtests.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/207)

### DIFF
--- a/vm/gce/utils.go
+++ b/vm/gce/utils.go
@@ -31,6 +31,9 @@ fi
 sudo chmod 777 /mnt/data1
 # sshguard can prevent frequent ssh connections to the same host. Disable it.
 sudo service sshguard stop
+# increase the default maximum number of open file descriptors. Load generators
+# running a lot of concurrent workers bump into this often.
+sudo sh -c 'echo "* - nofile 65536" > /etc/security/limits.d/10-roachprod-nofiles.conf'
 sudo touch /mnt/data1/.roachprod-initialized
 `
 


benesch:
Not that it matters much, but convention is to put automatically generated /etc config into the .d directory. So rather than appending to /etc/limits.conf, you’d write to /etc/limits.d/10-roachprod-nofiles.conf. (If you do this, double check that I’ve gotten the naming convention right.) That way your script is idempotent.

nvanbenschoten:
> You've verified this works manually, right? In my experience, there was always something else I needed to do.

Yes, I spun up a cluster and verified that all nodes reported the new limit. When I was originally testing this, I found that I needed to start a new shell to see the effect after updating the conf file. Maybe that's why it appeared like there was another step.

> Not that it matters much, but convention is to put automatically generated /etc config into the .d directory. So rather than appending to /etc/limits.conf, you’d write to /etc/limits.d/10-roachprod-nofiles.conf. (If you do this, double check that I’ve gotten the naming convention right.) That way your script is idempotent.

TIL! For those who are interested, see https://unix.stackexchange.com/questions/365622/how-to-raise-open-file-limit-etc-security-limits-d-vs-etc-security-limits-co for more.

PR #209, opened 2018-10-01
Increase default maximum open file descriptors for root user

nvanbenschoten:
The * in `/etc/security/limits.conf` does not apply to the root user. See https://stackoverflow.com/questions/38991351/why-in-etc-security-limits-conf-doesnt-include-root-user.

This means that previous attempts to increase the default max open file descriptors limit did not have an effect on nightly roachtest runs triggered by teamcity, which are run by root.

This commit increases the fd limit for the root user.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/209)


tschottdorf:
Nevermind, missed the second line.

PR #210, opened 2018-10-08
Add a retry-loop for retrieving internal IP addresses

petermattis:
We saw one occurrence where `GetInternalIP` returned an empty address
without an error. Add a retry loop similar to the one below for
`ssh-keyscan`.

Fixes github.com/cockroachdb/cockroach#31041

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/210)


petermattis:
🤞 that this fixes github.com/cockroachdb/cockroach#31041.

PR #212, opened 2018-10-11
Bump GCE Ubuntu image

tschottdorf:
See the deprecation warning in https://github.com/cockroachdb/cockroach/issues/28921#issuecomment-428899292

We should probably move to 18.04, but there's likely fallout and I don't
want to do it right now.


tschottdorf:
PS I have some hope that this fixes the apt-get failures because maybe it uses a better mirror. Not sure if that's at all true, though. Probably not.

petermattis:
LGTM

PR #215, opened 2018-10-16
Add `-i id_rsa` to the scp command as well

petermattis:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/215)


petermattis:
FYI, I had looked for all of the `ssh` invocations when reviewing your PR. I completely forgot to look for `scp`. Pretty sure this is the cause of this failure: https://teamcity.cockroachdb.com/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=969760&_focus=2476

I'm going to merge this so I can kick off another test run.

benesch:
D’oh! Good catch. We should really centralize all the SSH flag stuff somewhere.

PR #218, opened 2018-10-18
Add roachprod reformat xfs

tbg:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/218)

### DIFF
--- a/main.go
+++ b/main.go
@@ -887,6 +887,7 @@ WARNING: Reformatting will delete all existing data in the cluster.
 Filesystem options:
   ext4
   zfs
+  xfs
 
 When running with ZFS, you can create a snapshot of the filesystem's current
 state using the 'zfs snapshot' command:
@@ -916,6 +917,8 @@ the 'zfs rollback' command:
 			fsCmd = `sudo zpool create -f data1 -m /mnt/data1 /dev/sdb`
 		case "ext4":
 			fsCmd = `sudo mkfs.ext4 -F /dev/sdb && sudo mount -o discard,defaults /dev/sdb /mnt/data1`
+		case "xfs":
+			fsCmd = `sudo apt-get update && sudo apt-get install xfsprogs && sudo mkfs.xfs -f -n ftype=1 /dev/sdb && sudo mount -o inode64 /dev/sdb /mnt/data1`
 		default:
 			return fmt.Errorf("unknown filesystem %q", fs)
 		}


petermattis (approved):
:lgtm:

Cc @andreimatei so you can be aware of this.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/218#-:-LP6pw6b9I0vSdgkYNy7:bsp8qzh)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @tschottdorf)

___
*[main.go, line 921 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/218#-LP6pfl334xheoiYiJVP:-LP6pfl334xheoiYiJVQ:ble33ly) ([raw file](https://github.com/cockroachdb/roachprod/blob/edffd3c35f46da89f051c73bafa72b3d6d0c0cb5/main.go#L921)):*
> ```Go
> 			fsCmd = `sudo mkfs.ext4 -F /dev/sdb && sudo mount -o discard,defaults /dev/sdb /mnt/data1`
> 		case "xfs":
> 			fsCmd = `sudo apt-get update && sudo apt-get install xfsprogs && sudo mkfs.xfs -f -n ftype=1 /dev/sdb && sudo mount -o inode64 /dev/sdb /mnt/data1`
> ```

Perhaps separate out the installation into an `install xfs` target as is done for `zfs`.

andreimatei (commented):
how hard would it be to make `roachprod create` support giving you a desired filesystem directly, without running `roachprod reformat` ?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/218#-:-LP6vRKo3PqvTt3ZPjDb:b-2mih5q)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @tschottdorf)

benesch:
> how hard would it be to make roachprod create support giving you a desired filesystem directly, without running roachprod reformat ?

Easy enough. Just plumb an option through the create code path that internally calls reformatCmd. What's the use case, though? Just convenience?

`roachprod reformat` is kinda slow because it has to install a bunch of reps.

andreimatei (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/218)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @tschottdorf)

___
*[main.go, line 921 at r1](https://reviewable.io/reviews/cockroachdb/roachprod/218#-LP6vioO154BV-IWfyvo:-LP6vioO154BV-IWfyvp:b-8lfsqn) ([raw file](https://github.com/cockroachdb/roachprod/blob/edffd3c35f46da89f051c73bafa72b3d6d0c0cb5/main.go#L921)):*
> ```Go
> 			fsCmd = `sudo mkfs.ext4 -F /dev/sdb && sudo mount -o discard,defaults /dev/sdb /mnt/data1`
> 		case "xfs":
> 			fsCmd = `sudo apt-get update && sudo apt-get install xfsprogs && sudo mkfs.xfs -f -n ftype=1 /dev/sdb && sudo mount -o inode64 /dev/sdb /mnt/data1`
> ```

I'm curious - is /`dev/sdb` mounted anywhere at this point, and is anything mounted at`/mnt/data1` at this point? 
If the answer is yes to any, does this `mkfs` and `mount` still just work?

andreimatei (commented):
One can call it "just convenience"; I'd call it "a consistent API". When I create a cluster, I ask for the number of CPUs and machines and whatever else I want. I don't create a random cluster and then start terra-forming it into what I want little by little.
This is not to say that I'm against the `reformat` command. It's legit to want to reformat an existing cluster too. Same for remounting ext4 with nobarrier.

Just nerd-sniping; I'm interested in the topic because interactions with roachtest cluster reuse are on my mind.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/218#-:-LP6wM8SCd31lKoHcWZo:bqg535x)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @tschottdorf)

benesch:
> When I create a cluster, I ask for the number of CPUs and machines and whatever else I want. I don't create a random cluster and then start terra-forming it into what I want little by little.

By that logic, you should also request all the dependencies that you want installed at cluster creation time. You need Java for Hibernate? Then you need `roachprod create --software=openjdk`. 

Anyway, in the current world we live in, it's quick and cheap to flip between fs types, so we might as well allow it so that we can schedule ZFS and ext4 tests on the same VM resources. I don't think it's a bad thing that our roachprod API separates things that are slow (cluster creation) from things that are fast (installing deps, reformatting the FS).

andreimatei (commented):
> By that logic, you should also request all the dependencies that you want installed at cluster creation time. You need Java for Hibernate? Then you need roachprod create --software=openjdk.

Yeah, and that command looks exactly right to me :). It would lead to easily sharable command lines that people can copy paste for running various workloads by hand. If the cluster is useless to me without `openjdk` installed, then I might as well ask for it when creating the cluster. I don't see something being slow or fast as needing to have a particular impact on the design of this API.

> so we might as well allow it so that we can schedule ZFS and ext4 tests on the same VM resources.

yeah, that's why I think the `reformat` command should **also** exist.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/218#-:-LP6z9lX20BsRM0YemJk:b3q2lsq)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @tschottdorf)

benesch:
> Yeah, and that command looks exactly right to me :). It would lead to easily sharable command lines that people can copy paste for running various workloads by hand. If the cluster is useless to me without openjdk installed, then I might as well ask for it when creating the cluster. I don't see something being slow or fast as needing to have a particular impact on the design of this API.

I think roachprod is the wrong tool for this job. roachprod is a building block for creating and manipulating clusters. Something higher level should be responsible for getting them into the state that you need. I think roachtest is probably the right place.

andreimatei:
> I think roachprod is the wrong tool for this job. roachprod is a building block for creating and manipulating clusters. Something higher level should be responsible for getting them into the state that you need. I think roachtest is probably the right place.

For the sake of the argument, you'd have a point if `roachprod` was a tool that was not preoccupied with installing software or reformatting drives (like the gce cli is; that's a "building block"). That's when the "do one thing and do it well" philosophy holds water - when it helps separate different concerns to different code modules. But `roachprod` is very much preoccupied with software and filesystems, so it might as well present the best API it can around them.

andreimatei:
roachprod has moved to the main repo

tbg:
I'm in no rush on this one, so going to just close.

PR #221, opened 2018-10-19
Install and configure chrony on AWS clusters

petermattis:
Fixes cockroachdb/cockroach#31577

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/221)

### DIFF
--- a/vm/aws/support.go
+++ b/vm/aws/support.go
@@ -36,6 +36,12 @@ if [ "${disknum}" -eq "0" ]; then
   mkdir -p /mnt/data1
   chmod 777 /mnt/data1
 fi
+
+sudo apt-get update
+sudo apt-get install -qy chrony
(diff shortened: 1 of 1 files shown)


benesch:
What else is in the file? If not much, we can just entirely overwrite it.

petermattis:
It contains quite a bit:

```
~ grep -v '^#' /etc/chrony/chrony.conf | grep '.'
pool 2.debian.pool.ntp.org offline iburst
keyfile /etc/chrony/chrony.keys
commandkey 1
driftfile /var/lib/chrony/chrony.drift
log tracking measurements statistics
logdir /var/log/chrony
maxupdateskew 100.0
dumponexit
dumpdir /var/lib/chrony
logchange 0.5
hwclockfile /etc/adjtime
rtcsync
server 169.254.169.123 prefer iburst
```

benesch:
Huh, ok. I'm happy with the code as it stands. If the default changes it'll probably change to make the AWS time sync server the default, so seems safe enough.

PR #224, opened 2018-10-23
Disable host key checking in non-interactive sessions

tschottdorf:
IP reuse is frequent in our GCE test clusters.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/224)

### DIFF
--- a/install/session.go
+++ b/install/session.go
@@ -32,6 +32,7 @@ type remoteSession struct {
 func newRemoteSession(user, host string) (*remoteSession, error) {
 	args := []string{
 		user + "@" + host,
+		"-o", "UserKnownHostsFile=/dev/null",
 		"-o", "StrictHostKeyChecking=no",
 	}
 	args = append(args, sshAuthArgs()...)


tschottdorf:
Ah, hmm, maybe this message doesn't actually cause problem except that it gets displayed. (I was running a `grep` across as 12 node cluster).

tschottdorf:
Yeah, looks like it. I assume we're not going to merge this then.

tschottdorf:
Actually the warning message without this is pretty distracting.

install/session.go
```diff
@@ -34,6 +34,7 @@ func newRemoteSession(user, host string) (*remoteSession, error) {
 		user+"@"+host,
 		"-i", filepath.Join(config.OSUser.HomeDir, ".ssh", "id_rsa"),
 		"-i", filepath.Join(config.OSUser.HomeDir, ".ssh", "google_compute_engine"),
+		"-o" ,"UserKnownHostsFile=/dev/null",
```
petermattis: Weird spacing around the comma.

PR #226, opened 2018-10-23
Force chrony to synchronize

petermattis:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/226)

### DIFF
--- a/vm/aws/support.go
+++ b/vm/aws/support.go
@@ -40,7 +40,9 @@ fi
 sudo apt-get update
 sudo apt-get install -qy chrony
 echo -e "\nserver 169.254.169.123 prefer iburst" | sudo tee -a /etc/chrony/chrony.conf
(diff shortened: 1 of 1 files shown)


petermattis:
Unfortunately, this doesn't seem to work properly. After the `chronyc -a makestep`, the clocks are still not synchronized for a bit. Sending this out in the hopes that it nerd snipes one of you into tracking down why.

mberhault:
I don't see `ntp` being disabled anywhere. Isn't it installed by default?

petermattis:
@mberhault:

```
~ ps axuww | grep ntp
ubuntu    3632  0.0  0.0  12944   968 pts/0    S+   13:51   0:00 grep --color=auto ntp
```

petermattis:
I think this is working now. The key was the use of `waitsync`. The output from that indicates that `chronyd` is often taking a few seconds to perform the synchronization.

petermattis:
Ping. I'd like to merge this before tonight's nightly roachtest run.

PR #230, opened 2018-10-29
Add instrumentation to debug `roachprod stop` problem

petermattis:
See github.com/cockroachdb/cockroach#31586. `roachprod start` preceded
by `roachprod stop --wait` sometimes fails because a file lock has not
yet been released. Add instrumentation to `roachprod start` to list the
roachprod processes and the file locks in order to provide some insight
as to what is going on.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/230)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -134,15 +134,19 @@ func (c *SyncedCluster) Stop(sig int, wait bool) {
 
 		// NB: the awkward-looking `awk` invocation serves to avoid having the
 		// awk process match its own output from `ps`.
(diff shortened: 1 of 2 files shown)


petermattis:
I suppose I could add this to `stop` instead. My instinct was to add this instrumentation immediately before the failing `start` command.

petermattis:
Good point about dumping all of the processes. I've switched this up so that we dump the process list in both `roachprod stop` and `roachprod start`. And the output is placed in a new `logs/roachprod.log` file. PTAL.

PR #234, opened 2018-11-03
Revert "Rework stop-and-wait"

petermattis:
This reverts commit 777f15eeefa88b578853b565459b416af2fff228.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/234)

### DIFF
--- a/install/cluster_synced.go
+++ b/install/cluster_synced.go
@@ -132,32 +132,40 @@ func (c *SyncedCluster) Stop(sig int, wait bool) {
 		}
 		defer session.Close()
 
-		var waitCmd string
-		if wait {
-			waitCmd = `
-  for pid in ${pids}; do
-    while kill -0 ${pid}; do
-      sleep 1
-    done
-  done
-`
-		}
-
 		// NB: the awkward-looking `awk` invocation serves to avoid having the
 		// awk process match its own output from `ps`.
 		cmd := fmt.Sprintf(`
 mkdir -p logs
 echo ">>> roachprod stop: $(date)" >> %[1]s/roachprod.log
 ps axeww -o pid -o command >> %[1]s/roachprod.log
+count=0
+while :; do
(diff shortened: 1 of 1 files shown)


petermattis:
Looks like this broke local acceptance tests. I don't see why and can't reproduce the breakage locally. I'm going to merge this without review with 🤞 that this fixes the problem.

tschottdorf:
Thanks, was just noticing this.

tschottdorf:
(I also don't get this behavior on osx, at least not deterministically, and it seems deterministic on the TC agents)

petermattis:
Yep, reverting seems to have fixed the problem. Not sure what is going on. I'm unable to reproduce a failure on osx or on the builder on osx. I'll investigate more tomorrow.

petermattis:
I'm not sure why https://github.com/cockroachdb/roachprod/commit/777f15eeefa88b578853b565459b416af2fff228 broke local acceptance tests on CI. So far I'm unable to reproduce any breakage. I've attempted both on osx and within the builder image on osx.

See https://teamcity.cockroachdb.com/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=1000560&_focus=1491 for an example failure. It appears that `roachprod stop` is never returning. I'm speculating that something isn't working right with `kill -0`. @benesch Any thoughts?

benesch:
No, I'm at a loss. All I have is to suggest that we move roachprod into cockroachdb/cockroach so that they're versioned together, both to make debugging this easier and to prevent changes to roachprod from breaking unrelated cockroach PRs.

petermattis:
@benesch Did your suggestion? The roachtests are failing in the PR: https://teamcity.cockroachdb.com/viewLog.html?buildId=1009689&buildTypeId=Cockroach_UnitTests. Now to track down why.

PR #235, opened 2018-11-09
Format multiple SSDs as single RAID 0 volume on AWS

nvanbenschoten:
This commit adjusts the AWS startup script to format and mount
multiple SSDs as a single volume using RAID 0. It follows the
instructions from
https://cloud.google.com/compute/docs/disks/local-ssd#formatmultiple.

This is important for AWS because many instance types come with
multiple instance store volumes by default. For instance, `m5d.4xlarge`
comes with 2 300GB local SSDs. The full list of defaults is in
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html.

We'll probably want similar logic in GCE's startup script, although
that's not as important because GCE doesn't provide multiple SSDs
by default for any instance types.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/235)

### DIFF
--- a/vm/aws/support.go
+++ b/vm/aws/support.go
@@ -16,28 +16,41 @@ import (
 // https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html
 const awsStartupScript = `#!/usr/bin/env bash
 set -x
-disknum=0
+sudo apt-get update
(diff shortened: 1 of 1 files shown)


nvanbenschoten:
TFTRs!

PR #237, opened 2018-11-13
Introduce a --local-ssd-no-ext4-barrier flag for create

andreimatei:
Use of nobarrier is common for roachprod clusters. They might even
become the default for roachtest clusters. This patch bring first-class
support for them into roachprod, which learns to mount a drive
accrdingly.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/roachprod/237)


petermattis:
I'm going to hold off on reviewing this until https://github.com/cockroachdb/cockroach/pull/32267 is in. You'll have to move the PR over there. Sorry for the inconvenience.

